PR #129, opened 2020-05-27
discover: optional hint for probing

jwkohnen:
Add an functional option that parametrizes and retrieves per callback an opaque hint for the most recent successful probe to use.  That hint is a simple `int` that indexes into the static slice of probe functions.  The probe indices are implementation detail and thus the `int` is opaque.  If the a persisted hint gets out of sync with the implementation or if the machine behind the coordinate that is probed changes to a different model, the hinting either just makes the probing sub-optimal (hint to a non-matching probe) or reverts to default behaviour (hint out of bounds).

The functional option pattern is used in order to keep the default behaviour (no hinting) simple and to keep the published function signature of `discover.ScanAndCoonect()` backwards compatible.

### DIFF
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -3,18 +3,45 @@ package discover
 import (
 	"os"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/bmc-toolbox/bmclib/errors"
 	"github.com/bmc-toolbox/bmclib/internal/httpclient"
+	_ "github.com/bmc-toolbox/bmclib/logging" // this make possible to setup logging and properties at any stage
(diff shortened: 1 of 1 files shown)


jwkohnen:
Note about a risk that just came to my mind:  The order in which the probing happens right now, may unintentionally and unknowingly be important.  Probe `n` may depend on the fact that probe `n-i` is a negative match, thus probing `n` before `n-i` (caused by a bogus hint) might produce a false positive match. 

We'd have to double check the specificity of the probes.

jwkohnen:
closing in favour of #130

PR #135, opened 2020-09-23
Generic logging interface

jacobweinstock:
moving the logging to a generic interface so clients can choose the logger to use. 

- This implementation takes in an optional logger and if none is provided by the client then we default to using logrus like exists now. There is an example implementation in examples/main.go

- the logging interface used is https://github.com/go-logr/logr. the docs are good and the interface is really simple. numerous implementations already exist as well.


joelrebel (approved):
looks great! lets tag this as a 0.4.x since it changes various func signatures

PR #144, opened 2020-11-05
Add distinct http port to new iDrac9

jacobweinstock:
Add distinct HTTP port to new iDrac9.
Adding a distinct http port makes testing easier.
mock ssh and http servers can be spun up and passed in.


jacobweinstock:
code merged into #143

PR #145, opened 2020-11-11
Add generic provider

jacobweinstock:
The generic provider tries all interface actions via `ipmitool` as a catch-all for vendors that aren't implemented yet. Not all functionality is available yet. Functionality in this PR: 
- power on
- power off
- power status
- setting the next boot device

This PR also puts out another way we could do the devices interface. smaller interfaces with stronger abstraction than the existing interfaces. It makes adding new providers easier as we don't have to implement so many interface methods but allows for expanding interface functionality without breaking the interface.


joelrebel (approved):
👍

jgavinray (approved):
LGTM

jacobweinstock:
duplicate of https://github.com/bmc-toolbox/bmclib/pull/157 but https://github.com/bmc-toolbox/bmclib/pull/157 uses the use interfaces

PR #146, opened 2020-11-12
Handle ssh keyboard-interactive auth

abdrabo:
On iDRAC9 4.22.00.53 ssh started to use keyboard-interactive method
instead of password method (on iDRAC9 4.10.10.10) and answering with
nothing causes the auth to fail.

There is a similar discussion on the forum in:
https://www.dell.com/community/Systems-Management-General/iDRAC8-2-70-70-70-SSH-keyboard-interactive-authentication/td-p/7427565


jacobweinstock (approved):
seems reasonable to me!

PR #147, opened 2020-11-13
Add Dell PowerEdge R740xd to probe

nnuss:
Enable discovery of Dell PowerEdge R740xd machines using idrac9 so we can interact with them.


abdrabo (approved):
LGTM

PR #150, opened 2020-11-23
Fix idrac9 session exhaustion:

jacobweinstock:
Running probe multiple times exhausts the login
sessions for idrac9 as the idrac.New would open a
session to the idrac but not close it. I moved the
login to Users as this was the original reason to
have the login in New.


joelrebel (approved):
thanks!

PR #151, opened 2020-11-23
Log err instead of err.Error

abdrabo:
bmcbutler started to crash after updating bmclib module to v0.4.4. This
happens in cases like `if err != nil || statusCode != 200 { ... }`

Examples of stacktraces:

[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb2b9ae]

goroutine 140743 [running]:
github.com/bmc-toolbox/bmclib/providers/hp/ilo.(*Ilo).User(0xc004e56480, 0xc000112580, 0x3, 0x3, 0x1, 0xc005464c30)
    bmcbutler/vendor/github.com/bmc-toolbox/bmclib/providers/hp/ilo/configure.go:205 +0xa6e
github.com/bmc-toolbox/bmcbutler/pkg/butler/configure.(*Bmc).Apply(0xc00617f540)

---

[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb3007b]

goroutine 29007 [running]:
github.com/bmc-toolbox/bmclib/providers/hp/ilo.(*Ilo).Ntp(0xc0068ca000, 0xc004246c80, 0xc005820398, 0x1)
    bmcbutler/vendor/github.com/bmc-toolbox/bmclib/providers/hp/ilo/configure.go:433 +0xe7b
github.com/bmc-toolbox/bmcbutler/pkg/butler/configure.(*Bmc).Apply(0xc0001dba40)

---

[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb2d1b5]

goroutine 511214 [running]:
github.com/bmc-toolbox/bmclib/providers/hp/ilo.(*Ilo).Syslog(0xc001d46180, 0xc0005f8700, 0xc0008c4398, 0x1)
    bmcbutler/vendor/github.com/bmc-toolbox/bmclib/providers/hp/ilo/configure.go:279 +0x795
github.com/bmc-toolbox/bmcbutler/pkg/butler/configure.(*Bmc).Apply(0xc0007685a0)

### DIFF
--- a/providers/dell/idrac8/configure.go
+++ b/providers/dell/idrac8/configure.go
@@ -90,7 +90,7 @@ func (i *IDrac8) User(cfgUsers []*cfgresources.User) (err error) {
 			"step", "applyUserParams",
 			"IP", i.ip,
 			"Model", i.HardwareType(),
-			"Error", err.Error(),
+			"Error", err,
 		)
 		return err
 	}
@@ -103,7 +103,7 @@ func (i *IDrac8) User(cfgUsers []*cfgresources.User) (err error) {
 			"step", "applyUserParams",
 			"IP", i.ip,
 			"Model", i.HardwareType(),
-			"Error", err.Error(),
+			"Error", err,
 		)
 		return err
 	}
@@ -124,7 +124,7 @@ func (i *IDrac8) User(cfgUsers []*cfgresources.User) (err error) {
 						"Model", i.HardwareType(),
 						"step", helper.WhosCalling(),
 						"User", cfgUser.Name,
-						"Error", err.Error(),
+						"Error", err,
 					)
 					continue
 				}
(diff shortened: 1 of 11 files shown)


jacobweinstock:
I think the underlying issue here is not the `err.Error()` but the logic of the `if err != nil || statusCode != 200`. I would guess that the case you're running into is that `err` is `nil` but `statusCode` is not `200`, thus the `nil pointer dereference ` panic. 
With this PR would lose the value of `err` if we pass the `err` and not a string. Here's an example: https://goplay.tools/snippet/jnOJxV2Gz-2 
I'm inclined to suggest we change the logic or check for nil on the error before using it in a logline. Thoughts?

jacobweinstock:
@abdrabo , here's another idea. we create a small function that takes in the err and returns either `err.Error()` or `""`. Something like this: https://goplay.tools/snippet/aJtrPUIXVhu 
just thoughts, let me know what you think.

abdrabo:
I like the ErrStringOrEmpty solution. It is simple and it requires less changes to the code.

abdrabo:
I am closing this PR and I will open another one

PR #154, opened 2020-11-24
Avoid crashing while logging if err is nil

abdrabo:
There are many places in the code like:

`if err != nil || statusCode != 200 { ... }`

which cause runtime errors because err can be nil then err.Error() is
invoked.

Examples of stacktraces:

[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb2b9ae]

goroutine 140743 [running]:
github.com/bmc-toolbox/bmclib/providers/hp/ilo.(*Ilo).User(0xc004e56480, 0xc000112580, 0x3, 0x3, 0x1, 0xc005464c30)
bmcbutler/vendor/github.com/bmc-toolbox/bmclib/providers/hp/ilo/configure.go:205 +0xa6e
github.com/bmc-toolbox/bmcbutler/pkg/butler/configure.(*Bmc).Apply(0xc00617f540)

---

[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb3007b]

goroutine 29007 [running]:
github.com/bmc-toolbox/bmclib/providers/hp/ilo.(*Ilo).Ntp(0xc0068ca000, 0xc004246c80, 0xc005820398, 0x1)
bmcbutler/vendor/github.com/bmc-toolbox/bmclib/providers/hp/ilo/configure.go:433 +0xe7b
github.com/bmc-toolbox/bmcbutler/pkg/butler/configure.(*Bmc).Apply(0xc0001dba40)

---

[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb2d1b5]

goroutine 511214 [running]:
github.com/bmc-toolbox/bmclib/providers/hp/ilo.(*Ilo).Syslog(0xc001d46180, 0xc0005f8700, 0xc0008c4398, 0x1)
bmcbutler/vendor/github.com/bmc-toolbox/bmclib/providers/hp/ilo/configure.go:279 +0x795
github.com/bmc-toolbox/bmcbutler/pkg/butler/configure.(*Bmc).Apply(0xc0007685a0)

### DIFF
--- a/providers/supermicro/supermicrox/supermicrox.go
+++ b/providers/supermicro/supermicrox/supermicrox.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/bmc-toolbox/bmclib/devices"
 	"github.com/bmc-toolbox/bmclib/errors"
+	"github.com/bmc-toolbox/bmclib/internal"
 	"github.com/bmc-toolbox/bmclib/internal/httpclient"
 	"github.com/go-logr/logr"
 
@@ -247,7 +248,7 @@ func (s *SupermicroX) HardwareType() (model string) {
 	m, err := s.Model()
 	if err != nil {
 		// Here is your sin
-		s.log.V(1).Info("error getting hardwaretype", "err", err.Error())
+		s.log.V(1).Info("error getting hardwaretype", "err", internal.ErrStringOrEmpty(err))
 		return model
 	}
 	if strings.Contains(strings.ToLower(m), X10) {
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -6,3 +6,10 @@ import "unicode"
 func IsntLetterOrNumber(c rune) bool {
 	return !unicode.IsLetter(c) && !unicode.IsNumber(c)
 }
+
+func ErrStringOrEmpty(err error) string {
+	if err == nil {
+		return ""
+	}
+	return err.Error()
+}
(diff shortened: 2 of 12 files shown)


abdrabo:
There is a related discussion in https://github.com/bmc-toolbox/bmclib/pull/151

providers/supermicro/supermicrox/supermicrox.go
```diff
@@ -247,7 +248,7 @@ func (s *SupermicroX) HardwareType() (model string) {
 	m, err := s.Model()
 	if err != nil {
 		// Here is your sin
-		s.log.V(1).Info("error getting hardwaretype", "err", err.Error())
+		s.log.V(1).Info("error getting hardwaretype", "err", internal.ErrStringOrEmpty(err))
```
jacobweinstock, line 251: probably only need the `ErrStringOrEmpty` for loglines that check `if err != nil || ...`. other than that looks good, thanks!

jacobweinstock (approved):
just one comment, but won't let it hold up the PR. Thanks!

providers/supermicro/supermicrox/supermicrox.go
abdrabo: I changed it everywhere to be more consistent but I can change it only for `if err != nil || ...` if you think it is better that way.

jacobweinstock: up to you. it's not needed in some places, but I don't think it hurts either.

Issue #148, opened 2020-11-17
new design proposal 


jacobweinstock:
# BMCLIB Design Doc

## Context and Scope
There are many ways to interact with a BMC. IPMI, Redfish, web API, RACADM, SSH, etc. Not all interaction types are implemented for every vendor. Even if they were, there is no functionality to try multiple interaction types for a request.

The main entrypoint into BMCLIB is `discover.ScanAndConnect`. This function returns one of two very large [interfaces](#Existing-interface). The user then must type assert or switch on the interface to start using the functionality.

The aim of this new design is to
- create stronger abstractions through smaller interfaces
- enhance the success of BMC interactions by trying requests through multiple interaction types
- allow easier extension of functionality, vendors and interaction types.

## Goals and Non-Goals
**Goals:**

- Create stronger abstraction through a smaller [interfaces](#Proposed-interfaces)
- Allow provider implementations the ability to grow in implemented functionality without breaking Users
- Increase the success of BMC requests by trying multiple interaction types
- Increase simplicity of adding providers

**Non-Goals:**
- Change any existing provider implementation details
- Make API breaking changes

# New Design

## Overview
Some core tenets of this new design are
- Small interfaces
- Implementations are only run if they satisfy the interface requirements
- Focus on accomplishing a desired BMC interaction not how its done or via what mechanism (ie redfish, ipmitool, web api, etc)
- All interface functions will try multiple implementations (ipmitool, web api, redfish, ssh, etc) until success or failure by receiving a parameter that is a slice of PowerStateSetter interfaces, shown below.
`func SetPowerState(ctx context.Context, s string, p []PowerStateSetter) (bool, error)`

## Details
The existing interface methods are all pretty good. The idea here is to split them out into smaller [interfaces](#Proposed-interfaces). This new design can be introduced without breaking existing functionality. Existing functionality would plugin to this and users could choose to use either or both. The `ScanAndConnect` function would be integrated into the Client box below under the “provider registry helper”.  See the code [here](https://github.com/jacobweinstock/bmclib/blob/new-design-proposal/client_test.go) for working code of this design.
Context Diagram
<img width="925" alt="bmclib-redesign-flow" src="https://user-images.githubusercontent.com/12081036/99327968-59837880-2838-11eb-8a74-7535fe3cfacf.png">

## Code Snippets
**User Package**
```go
package main
 
import (
   "context"
   "log"
 
   "github.com/bmc-toolbox/bmclib"
)
 
func main() {
[...]

PR #155, opened 2020-11-24
New interfaces proposal

jacobweinstock:
Proposal for new interfaces to get us started with the new design. https://github.com/bmc-toolbox/bmclib/issues/148


bmc/power.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+// PowerSetter sets the power state of a BMC
+type PowerSetter interface {
+	PowerSet(ctx context.Context, state string) (ok bool, err error)
```
joelrebel, line 13: Since the `state` string is ambiguous, would it be worth to have pre defined Power state Consts/types, different vendor BMCs could return their own string for Power state, would it be useful if the the provider methods could return the predefined consts?

jacobweinstock: I believe there are trade-offs to using a custom type there. With a custom type, the implementors are now coupled to the interface package. With a string, the implementing package and this interface package aren't coupled. The other side of the trade-off is exactly what you said.

jacobweinstock: some middle ground could be some explicit comments on the interfaces to say something like: implementors should return one of these values. just one idea. I'm sure there are other options out there.

jacobweinstock: add some comments about this idea. let me know what you think?

joelrebel (approved):
🚀

PR #157, opened 2020-11-30
Add ipmitool implementation

jacobweinstock:
Add ipmitool implementation of new design interfaces. power, boot device, bmc reset and user read have been implemented. Added context to internal/ipmi calls.


joelrebel (approved):
LGTM

Issue #158, opened 2020-11-30
discover.ScanAndConnect errors if a BMC doesn't have a web endpoint on port 443


jacobweinstock:
`discover.ScanAndConnect` fails if the BMC host doesn't have a service to connect to on port 443.

```bash
Get "https://127.0.0.1/xmldata?item=all": dial tcp 127.0.0.1:443: connect: connection refused
```

How to reproduce:
```bash
# start a ipmi simulator
docker run -d -p 623:623/udp vaporio/ipmi-simulator

# edit examples/main.go with local bmc connection details
sed -i.bak 's/ip := "<bmc_ip>"/ip := "127.0.0.1"/g' examples/main.go
sed -i.bak 's/user := "user"/user := "ADMIN"/g' examples/main.go
sed -i.bak 's/pass := "password"/pass := "ADMIN"/g' examples/main.go
rm -rf example/main.go.bak

# run the main.go example
go run examples/main.go
FATA[0000] Get "https://127.0.0.1/xmldata?item=all": dial tcp 127.0.0.1:443: connect: connection refused 
exit status 1
```

Expected behavior:

I would expect that if a BMC doesn't server anything on port 443 then, that specific provider would fail out with a `errors.ErrDeviceNotMatched` or similar but not to fail the discovery process. Something possibly like this:

```bash
go run examples/main.go
FATA[0000] unable to identify the vendor 
```

PR #159, opened 2020-11-30
Fix premature probe failure

jacobweinstock:
`discover.ScanAndConnect` fails if the BMC host doesn't have a service to connect to on port 443. The expected behavior is to report an error and try the next provider. This PR will allow for this. Fixes https://github.com/bmc-toolbox/bmclib/issues/158

### DIFF
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -80,26 +80,23 @@ func ScanAndConnect(host string, username string, password string, options ...Op
 	for _, probeID := range order {
 		probeDevice := devices[probeID]
 
-		opts.Logger.V(1).Info("probing to identify device", "step", "ScanAndConnect", "host", host)
+		opts.Logger.V(1).Info("probing to identify device", "step", "ScanAndConnect", "host", host, "vendor", probeID)
 
 		bmcConnection, err := probeDevice(opts.Context, opts.Logger)
 
 		// if the device didn't match continue to probe
-		if err != nil && (err == errors.ErrDeviceNotMatched) {
+		if err != nil {
+			// log error if probe is not successful
+			opts.Logger.V(1).Info("probe failed", "step", "ScanAndConnect", "host", host, "vendor", probeID, "error", err)
 			continue
 		}
-
-		// at this point it could be a connection error or a errors.ErrUnsupportedHardware
-		if err != nil {
-			return nil, err
+		if hintErr := opts.HintCallback(probeID); hintErr != nil {
+			return nil, hintErr
 		}
-
-		if err := opts.HintCallback(probeID); err != nil {
-			return nil, err
-		}
-
 		// return a bmcConnection
-		return bmcConnection, nil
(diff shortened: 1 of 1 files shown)


discover/discover.go
```diff
@@ -80,26 +80,23 @@ func ScanAndConnect(host string, username string, password string, options ...Op
...
+		opts.Logger.V(1).Info("probing to identify device", "step", "ScanAndConnect", "host", host, "vendor", probeID)
 
-		bmcConnection, err := probeDevice(opts.Context, opts.Logger)
+		bmcConnection, probeErr := probeDevice(opts.Context, opts.Logger)
 
 		// if the device didn't match continue to probe
-		if err != nil && (err == errors.ErrDeviceNotMatched) {
+		if probeErr != nil {
+			// log error if probe is not successful
+			opts.Logger.V(1).Info("probe failed", "host", host, "vendor", probeID, "error", probeErr)
```
mikemrm, line 90: Should we include the step as well to keep the same structure as the previous log message?

discover/discover.go
```diff
@@ -80,26 +80,23 @@ func ScanAndConnect(host string, username string, password string, options ...Op
...
-
-		if err := opts.HintCallback(probeID); err != nil {
-			return nil, err
+		if hintErr := opts.HintCallback(probeID); hintErr != nil {
+			return nil, hintErr
 		}
-
 		// return a bmcConnection
-		return bmcConnection, nil
+		if bmcConnection != nil && probeErr == nil {
```
mikemrm, line 97: No need to check probeErr here since we `continue` if probeErr != nil on line 88 right?

discover/discover.go
```diff
@@ -80,26 +80,23 @@ func ScanAndConnect(host string, username string, password string, options ...Op
 	for _, probeID := range order {
 		probeDevice := devices[probeID]
 
-		opts.Logger.V(1).Info("probing to identify device", "step", "ScanAndConnect", "host", host)
+		opts.Logger.V(1).Info("probing to identify device", "step", "ScanAndConnect", "host", host, "vendor", probeID)
 
-		bmcConnection, err := probeDevice(opts.Context, opts.Logger)
+		bmcConnection, probeErr := probeDevice(opts.Context, opts.Logger)
```
mikemrm, line 85: I'm not sure it's necessary to have separate error variables, I believe it's typically common practice to re-use if there's no overlap.

mikemrm (approved):
lgtm

discover/discover.go
jacobweinstock: sure, will update

discover/discover.go
jacobweinstock: ah yes, good catch. thanks! will update

discover/discover.go
jacobweinstock: sure, will go back to `err`.

PR #160, opened 2020-11-30
Add client package

jacobweinstock:
Client package purpose is to be the primary interaction that a bmclib user would use. It interacts with the registry to initialize registered providers with auth and then has pass-through functions to call the interface executor functions.


joelrebel (approved):
Thanks for this, it'll help folks get going - you may want to link to this in, https://github.com/bmc-toolbox/bmclib/blob/master/doc.go

PR #161, opened 2020-12-01
Pass logger like context

jacobweinstock:
Pass a logging interface all the way down to the interface definitions allows us to add log entries in the interface executors/functions and is in line with the way context is passed around.


jacobweinstock:
closing in favor of https://github.com/bmc-toolbox/bmclib/pull/169

PR #167, opened 2020-12-04
Add basic OpenBMC support

zevweiss:
The only implements a small subset of the `devices.Bmc` interface, but hopefully it's enough for baseline PBnJ support?  Currently it's entirely redfish-based, though many of these operation could probably be alternatively implemented via IPMI or ssh (or perhaps the OpenBMC-specific HTTP API under `/xyz/openbmc_project/...`).

Also, I'm very much neither a Go expert nor real familiar with bmclib, so careful review is probably warranted to make sure I haven't done something egregiously foolish...

### DIFF
--- a/devices/constants.go
+++ b/devices/constants.go
@@ -16,6 +16,8 @@ const (
 	Cloudline = "Cloudline"
 	// Quanta is the contant to identify Quanta hardware
 	Quanta = "Quanta"
+
+	OpenBmc = "OpenBMC"
 	// Common is the constant of thinks we could use across multiple vendors
 	Common = "Common"
 
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -22,6 +22,7 @@ const (
 	ProbeM1000e        = "m1000e"
 	ProbeQuanta        = "quanta"
 	ProbeHpCl100       = "hpcl100"
+	ProbeOpenBmc       = "openbmc"
 )
 
 // ScanAndConnect will scan the bmc trying to learn the device type and return a working connection.
@@ -60,9 +61,11 @@ func ScanAndConnect(host string, username string, password string, options ...Op
 		ProbeM1000e:        probe.m1000e,
 		ProbeQuanta:        probe.quanta,
 		ProbeHpCl100:       probe.hpCl100,
+		ProbeOpenBmc:       probe.openBmc,
 	}
 
-	order := []string{ProbeHpIlo,
+	order := []string{ProbeOpenBmc,
+		ProbeHpIlo,
 		ProbeIdrac8,
 		ProbeIdrac9,
 		ProbeSupermicrox11,
--- a/discover/probe.go
+++ b/discover/probe.go
@@ -19,6 +19,7 @@ import (
 	"github.com/bmc-toolbox/bmclib/providers/hp"
 	"github.com/bmc-toolbox/bmclib/providers/hp/c7000"
 	"github.com/bmc-toolbox/bmclib/providers/hp/ilo"
+	"github.com/bmc-toolbox/bmclib/providers/openbmc"
 	"github.com/bmc-toolbox/bmclib/providers/supermicro/supermicrox"
 	"github.com/bmc-toolbox/bmclib/providers/supermicro/supermicrox11"
 	"github.com/go-logr/logr"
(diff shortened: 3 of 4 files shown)


jacobweinstock:
This is great! thanks, @zevweiss .

We've actually just recently added some new [interfaces](https://github.com/bmc-toolbox/bmclib/tree/master/bmc). Would you mind using those instead of the [old ones](https://github.com/bmc-toolbox/bmclib/blob/master/devices/interfaces.go)? The [ipmitool provider implementation](https://github.com/bmc-toolbox/bmclib/tree/master/providers/ipmitool) can be referenced as an example.

For some more context, the new design doc and discussion can be found [here](https://github.com/bmc-toolbox/bmclib/issues/148).

zevweiss:
Ah, hadn't noticed that those new interfaces were there -- thanks for the link to the background info.  Is there any existing tooling anywhere for testing out an in-progress implementation of the new interfaces?  For the current code (with the `devices.Bmc` interface) I used the code in `examples/main.go` to test it out a bit as I was putting it together; is there anything analogous for the new interfaces?  (I tried compiling the example code from the "User Package" snippet in https://github.com/bmc-toolbox/bmclib/issues/148#issue-744332957, but it errored out with "client.GenerateDefaultRegistry undefined" -- maybe that's something trivial/obvious to fix but I'm presently too much of a Go newb to know.)

jacobweinstock:
> Ah, hadn't noticed that those new interfaces were there -- thanks for the link to the background info. Is there any existing tooling anywhere for testing out an in-progress implementation of the new interfaces? For the current code (with the `devices.Bmc` interface) I used the code in `examples/main.go` to test it out a bit as I was putting it together; is there anything analogous for the new interfaces? (I tried compiling the example code from the "User Package" snippet in [#148 (comment)](https://github.com/bmc-toolbox/bmclib/issues/148#issue-744332957), but it errored out with "client.GenerateDefaultRegistry undefined" -- maybe that's something trivial/obvious to fix but I'm presently too much of a Go newb to know.)

Ah yeah, good question. [client_test.go](https://github.com/bmc-toolbox/bmclib/blob/master/client_test.go) would be the plae to look.

zevweiss:
Alright, thanks -- though having now seen a bit more of how that works, I'm a little uncertain about the appropriate way to proceed with this.  After some experimentation, it looks like the existing `ipmitool.Conn` interface's of `PowerStateGetter`, `PowerSetter`, `BootDeviceSetter`, `BMCResetter`, and `UserReader` are all already compatible with OpenBMC -- I suppose maybe we could supplement that with Redfish-based implementations or something, but given that would it be expected that pbnj.next's basic functionality should "just work" with OpenBMC?

(CC @joelrebel if he wants to chime in on anything here...)

jacobweinstock:
> Alright, thanks -- though having now seen a bit more of how that works, I'm a little uncertain about the appropriate way to proceed with this. After some experimentation, it looks like the existing `ipmitool.Conn` interface's implementations of `PowerStateGetter`, `PowerSetter`, `BootDeviceSetter`, `BMCResetter`, and `UserReader` are all already compatible with OpenBMC -- I suppose maybe we could supplement that with Redfish-based implementations or something, but given that would it be expected that pbnj.next's basic functionality should "just work" with OpenBMC with bmclib as it stands?
> 
> (CC @joelrebel if he wants to chime in on anything here...)

If you've already done the validation here with an OpenBMC then we should be good on the bmclib side. PBnJ.next does need to update to use these new interfaces, but after that OpenBMC's in PBnJ.next just work as you say. I think bmclib on its own will still need a generic Redfish implementation. Just my 2cents :)

zevweiss:
Ah, wait -- except I just remembered I forgot one thing.  The operations I listed do work against OpenBMC with current bmclib, plus one little hack I had forgotten I slipped in:

```diff
diff --git a/internal/ipmi/ipmi.go b/internal/ipmi/ipmi.go
index a5a383f70872..0e26bc457075 100644
--- a/internal/ipmi/ipmi.go
+++ b/internal/ipmi/ipmi.go
@@ -36,7 +36,7 @@ func New(username string, password string, host string) (ipmi *Ipmi, err error)
 }
 
 func (i *Ipmi) run(ctx context.Context, command []string) (output string, err error) {
-	ipmiArgs := []string{"-I", "lanplus", "-U", i.Username, "-E", "-N", "5"}
+	ipmiArgs := []string{"-I", "lanplus", "-U", i.Username, "-E", "-N", "5", "-C", "17"}
 	if strings.Contains(i.Host, ":") {
 		host, port, err := net.SplitHostPort(i.Host)
 		if err == nil {
```

OpenBMC's network-facing IPMI daemon only supports RAKP-HMAC-SHA256/HMAC-SHA256-128/AES-CBC-128 (I guess due to deprecating SHA1 a bit more aggressively than most).  Any suggestions on a clean way to shim that extra flag in in a sufficiently flexible/generic way?

zevweiss:
...actually, after some further investigation ([RHBZ](https://bugzilla.redhat.com/show_bug.cgi?id=1868637), [patch](https://github.com/ipmitool/ipmitool/commit/7772254b62826b894ca629df8c597030a98f4f72), [patch](https://github.com/ipmitool/ipmitool/commit/9452be87181a6e83cfcc768b3ed8321763db50e4)), it looks like ipmitool _should_ be able to figure this out on its own, but there hasn't been a release since the patches fixing that bug went in.  (I compiled its current master branch and verified that it does work without the `-C17` flag.)

Any thoughts on how to approach this?  We could perhaps sidestep the issue by just requiring a suitably-patched ipmitool as a dependency for OpenBMC support (in which case I don't think any changes to the bmclib code would be needed), though maybe that's an overly onerous requirement...

zevweiss:
After some discussion with @joelrebel, I think we can avoid the API complexities by just having the IPMI provider do a little discovery step of its own (i.e. try an innocuous operation with `-C17` and keep using that if it works).  Joel said he'd take care of implementing that, so I don't think there should really be much else needed for basic OpenBMC support.

PR #168, opened 2020-12-05
Add PreferProtocol functionality

jacobweinstock:
Allows ordering the registries based on a list of preferred protocol types.


joelrebel (approved):
🚀

PR #169, opened 2020-12-05
Add log interface to register function

jacobweinstock:
Client defined logging wasn't making it into provider implementations. This allows the client to define the logger or if not defined by the client we use `logging.DefaultLogger()`.


joelrebel (approved):
Thanks for this!

PR #171, opened 2020-12-06
Add ipmitool provider tests

jacobweinstock:
Add ipmitool provider tests.


joelrebel (approved):
TIL about bou.ke/monkey :)

PR #173, opened 2020-12-10
Fix long connection timeout in probes

jacobweinstock:
Found an issue with ScanAndConnect taking over 20 minutes to return if the BMC address wasn't reachable. Context passed into ScanAndConnect wasn't being used for HTTP calls. This commit looks to fix that by using the timeout/cancellation from the passed in user context.


joelrebel (approved):
👍

PR #174, opened 2020-12-12
Add context timeout/cancel checking

jacobweinstock:
If implementations don't or can't handle checking for context timeouts/cancels, this will check for them before each implementation is run in the executor functions.


joelrebel (approved):
👍

PR #175, opened 2020-12-13
Open close interfaces

jacobweinstock:
implementations that have stateful connections need an open/close method to connect and disconnect. this PR adds theses as main interface methods.

The client package is also updated to handle the open/close functionality. This way multiple calls can all use the same connection(s)/session(s). Another nice thing here is that you can use different registry filters for different calls. For example, if you want to set the next boot using redfish, but get users using ipmitool you can do that. Here's a snippet example, `client_test.go` also has a similar example.
```go
client := NewClient(host, port, user, pass, WithLogger(log))
err := client.Open(ctx)
if err != nil {
 panic(err)
}
defer client.Close(ctx)

client.Registry = client.Registry.PreferProtocol("redfish")
ok, err := client.SetBootDevice(ctx, "pxe")
if err != nil {
 panic(err)
}
fmt.Println(ok)

client.Registry = client.Registry.PreferProtocol("ipmitool")
users, err := client.ReadUsers(ctx)
if err != nil {
 panic(err)
}
fmt.Println(users)
```

### DIFF
--- a/bmc/connection.go
+++ b/bmc/connection.go
@@ -1,0 +1,107 @@
+package bmc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+// Opener interface for opening a connection to a BMC
+type Opener interface {
+	Open(ctx context.Context) error
+}
+
+// Closer interface for closing a connection to a BMC
+type Closer interface {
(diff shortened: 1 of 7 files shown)


joelrebel (approved):
good stuff, I like that we're able to re-use sessions and have the option to prefer the registry method 🚀

PR #187, opened 2021-01-15
Update registry package

jacobweinstock:
Replace the local registry package with https://github.com/jacobweinstock/registry. This makes the registry much simpler to interact with and removes the [magic](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html) of init functions and package-level variables. Makes registering providers more understandable through explicit registration.


joelrebel (approved):
👍

PR #192, opened 2021-01-25
Fix nil values in provider registration:

jacobweinstock:
fix for broken `NewClient` in v0.4.9.

In `NewClient`, calling `defaultClient.registerProviders()` before setting the `host`, `port`, `user`, and `pass` causes registering providers to not have `host`, `port`, `user`, and `pass` set and results in cascading errors.


jgavinray (approved):
:shipit:

PR #193, opened 2021-01-25
Remove print statement

jacobweinstock:
Remove debugging/erroneous print statement.


jgavinray (approved):
@jacobweinstock thanks for putting this together and for working so diligently to keep good code hygiene.

PR #194, opened 2021-02-02
Updating dependencies.

omar-m-othman:
For those who build using `-mod=vendor`.


jacobweinstock:
Hey @omar-m-othman , bmclib doesn't have a build. We don't ever run `go build`. I'm assuming you're running into a challenge when importing bmclib into a project that does do vendor builds? is that correct?

omar-m-othman:
> Hey @omar-m-othman , bmclib doesn't have a build. We don't ever run `go build`. I'm assuming you're running into a challenge when importing bmclib into a project that does do vendor builds? is that correct?

That's right... not exactly troubles actually, but I was going to build all our tools using the new `bmclib`... so I thought it was best to update all deps before that.

I've removed the vendor/ folder... just updated `go.mod` and `go.sum` for now.

jacobweinstock (approved):

PR #201, opened 2021-02-16
ScanAndConnectv2() based on the v2 bmclib interface

joelrebel:
The ScanAndConnectv2() helper wraps identifying the BMC, registering the provider and returning a bmclib v2 client


jacobweinstock (commented):
Hey @joelrebel sorry, I'm having trouble understanding the need for this PR, apologies. The existing code we have already does "probing" for all providers: https://github.com/bmc-toolbox/bmclib/blob/abe5feaea6b999313cb6ce7c01d85f6adaaef03e/client_test.go#L22
registering providers and creating a new client happened here:
https://github.com/bmc-toolbox/bmclib/blob/abe5feaea6b999313cb6ce7c01d85f6adaaef03e/client_test.go#L21

From what I can tell, this PR is looking to put these two lines together into a single call, for a different end-user experience? 

And when you want to use just a specific provider we already have
```go
cl.Registry.Drivers = cl.Registry.For("asrockrack")
```

Again, I'm sorry I'm not grasping the intent here. Would you mind walking me through it again?

joelrebel:
> Hey @joelrebel sorry, I'm having trouble understanding the need for this PR, apologies. The existing code we have already does "probing" for all providers:
> 
> https://github.com/bmc-toolbox/bmclib/blob/abe5feaea6b999313cb6ce7c01d85f6adaaef03e/client_test.go#L22
> 
> 
> registering providers and creating a new client happened here:
> https://github.com/bmc-toolbox/bmclib/blob/abe5feaea6b999313cb6ce7c01d85f6adaaef03e/client_test.go#L21
> 
> From what I can tell, this PR is looking to put these two lines together into a single call, for a different end-user experience?
> 
> And when you want to use just a specific provider we already have
> 
> ```go
> cl.Registry.Drivers = cl.Registry.For("asrockrack")
> ```
> 
> Again, I'm sorry I'm not grasping the intent here. Would you mind walking me through it again?

Thanks for pointing that out, my apologies, seems like I went on a pointless excursion there,
the Verifier interface is indeed a nicer approach.

PR #205, opened 2021-03-09
I got an index-out-of-range exception here!

omar-m-othman:


joelrebel:
Hey @omar-m-othman thanks for sending this fix in

can you fix a few aspects of this PR that will make it clearer,
a description which includes,
- details about the idrac firmware version this fails on
- updated test data and tests https://github.com/bmc-toolbox/bmclib/blob/ad9bfc46b08334222ee02c648ea4c29f22f1f014/providers/dell/idrac8/idrac8_test.go#L5641

omar-m-othman:
> Hey @omar-m-othman thanks for sending this fix in
> 
> can you fix a few aspects of this PR that will make it clearer,
> a description which includes,
> 
>     * details about the idrac firmware version this fails on
> 
>     * updated test data and tests https://github.com/bmc-toolbox/bmclib/blob/ad9bfc46b08334222ee02c648ea4c29f22f1f014/providers/dell/idrac8/idrac8_test.go#L5641

Hey @joelrebel,

I don't remember which one threw the exception (IDRAC8 or IDRAC9)... but I guess it doesn't really matter, since the fix is quite clear and simple. I've added the fix for both anyway.

I've actually failed to understand how I can add a different value to the test without adding another 5000 lines of code... so, can we merge this one and change the testing logic to allow for creating multiple BMCs later?

joelrebel (approved):

PR #208, opened 2021-03-14
ASRR firmware upload mem consumption fixes

joelrebel:
- Stream firmware payload without buffering
- Updates the `FirmwareBMCUpdate`, `FirmwareBIOSUpdate` interface methods


providers/asrockrack/helpers.go
```diff
@@ -80,30 +80,53 @@ func (a *ASRockRack) setFlashMode() error {
...
+		errCh <- form.Close()
+	}()
+
+	// multi-part content type
+	headers := map[string]string{
+		"Content-Type": form.FormDataContentType(),
+	}
 
 	// POST payload
 	_, statusCode, err := a.queryHTTPS(endpoint, "POST", pipeReader, headers, contentLength)
```
zevweiss, line 132: (Bear with me as I ask possibly kind of dumb questions here...)

If `queryHTTPS()` here takes an `io.Reader`, what are we achieving by spinning up a goroutine to shuffle data from `fwReader` through a pipe instead of just passing `fwReader` directly as the argument here?

zevweiss: I'm also a bit confused by the APIs involved; based on a brief test I just did with a copy of the `multipartSize()` code, it looks like the `contentLength` argument to `queryHTTPS()` is I think just a "metadata" length of some sort?  (e.g. dependent on the length of the filename itself, not the actual quantity of data in the file?)  But the actual quantity of data transferred in the POST request will ultimately have to be a function of the file size, right?

providers/asrockrack/asrockrack.go
```diff
@@ -148,7 +148,8 @@ func (a *ASRockRack) GetBMCVersion(ctx context.Context) (string, error) {
 // nolint: gocyclo
 // BMC firmware update is a multi step process
 // this method initiates the upgrade process and waits in a loop until the device has been upgraded
-func (a *ASRockRack) FirmwareUpdateBMC(ctx context.Context, fileReader io.Reader) error {
+// fileSize is required to setup the multipart form upload Content-Length
+func (a *ASRockRack) FirmwareUpdateBMC(ctx context.Context, fileReader io.Reader, fileSize int64) error {
```
zevweiss, line 152: Just as a general process comment, since the two things mentioned in the commit message here are as far as I can tell completely independent, it'd preferable to have them in separate commits.  (Makes review easier, and fine-grained git history is more useful for things like potential future reverts and bisecting.)

providers/asrockrack/helpers.go
```diff
@@ -83,8 +83,8 @@ func (a *ASRockRack) setFlashMode() error {
 func multipartSize(fieldname, filename string) int64 {
 	body := &bytes.Buffer{}
 	form := multipart.NewWriter(body)
-	form.CreateFormFile(fieldname, filename)
-	form.Close()
+	_, _ = form.CreateFormFile(fieldname, filename)
+	_ = form.Close()
```
zevweiss, line 87: This commit looks like something to (ideally) fold back into 968ea38dc70cdb76f608b1379b532c20978e949a via a fixup/rebase.

examples/v1/main.go
```diff
@@ -52,7 +52,12 @@ func main() {
 	}
 	defer fh.Close()
 
-	err = cl.UpdateBMCFirmware(ctx, fh)
+  fi, err := fh.Stat()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = cl.UpdateBMCFirmware(ctx, fh, fi.Size())
```
zevweiss, line 60: These are a bit more related as they both pertain to the `examples/...` code, but as with bd9fd26ceca67d363e3fc23b579a972d093a7af0 in an ideal world I think these changes would be two separate commits.

examples/v1/main.go
```diff
@@ -52,7 +52,7 @@ func main() {
 	}
 	defer fh.Close()
 
-  fi, err := fh.Stat()
+	fi, err := fh.Stat()
```
zevweiss, line 55: ...and likewise, it'd be nice for this to be folded back into b3d20797fd7fe77aa677f308022984313c2fe430.

examples/v1/main.go
zevweiss: ...though actually, looking at the change here it seems like this code wouldn't have compiled prior to this change?  In which case maybe this should have just been a part of 2055fe1d1d6de1bd3c47080ecd1197f2abc18d27...

providers/asrockrack/helpers.go
joelrebel: > (Bear with me as I ask possibly kind of dumb questions here...)
> 
> If `queryHTTPS()` here takes an `io.Reader`, what are we achieving by spinning up a goroutine to shuffle data from `fwReader` through a pipe instead of just passing `fwReader` directly as the argument here?

The bytes from the io.Reader (update bin file contents) need to be wrapped within a [multipart form](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) this is done by the goroutine which [sets up the form related fields](https://github.com/bmc-toolbox/bmclib/blob/968ea38dc70cdb76f608b1379b532c20978e949a/providers/asrockrack/helpers.go#L109) and writes to one end of the pipe passed into `queryHTTPS()`

joelrebel: > I'm also a bit confused by the APIs involved; based on a brief test I just did with a copy of the multipartSize() code, it looks like the contentLength argument to queryHTTPS() is I think just a "metadata" length of some sort? (e.g. dependent on the length of the filename itself, not the actual quantity of data in the file?) But the actual quantity of data transferred in the POST request will ultimately have to be a function of the file size, right?

The [multipartSize](https://github.com/bmc-toolbox/bmclib/blob/968ea38dc70cdb76f608b1379b532c20978e949a/providers/asrockrack/helpers.go#L83) method helps figure the `Content-Length` - which includes the `firmware file size` + `multipart form fields` that have to be sent as part of the Request header - before the entire firmware file is read from disk. The filesize cannot be inferred from the io.Reader - not until all the bytes are read, I suppose?

bmc/firmware.go
```diff
@@ -16,7 +16,7 @@ type BMCVersionGetter interface {
 
 // BMCFirmwareUpdater upgrades the BMC firmware
 type BMCFirmwareUpdater interface {
-	FirmwareUpdateBMC(ctx context.Context, fileReader io.Reader) (err error)
+	FirmwareUpdateBMC(ctx context.Context, fileReader io.Reader, fileSize int64) (err error)
```
jacobweinstock, line 19: if this generic enough for hardware that isn't ASRR?

jacobweinstock (commented):
just one small question from me. Approved from me my end, but will wait for Zev's comments to be addressed before signing off.

examples/v1/main.go
joelrebel: fixed, now is a part of 808952d2

examples/v1/main.go
joelrebel: squashed in 808952d2d3

bmc/firmware.go
joelrebel: Im not too happy with the func signature there, and it is subject to change - In another iteration, I'd prefer to just pass in the URL to the file and let the method wrap the rest.

For now this will work with Redfish on the Dells and SMC - since they both use multi-part form upload.

HPE requires a URL https://github.com/HewlettPackard/python-redfish-utility/blob/master/examples/Linux/Update_Firmware.sh#L38

providers/asrockrack/helpers.go
joelrebel: squashed into 7f3f530

examples/v1/main.go
zevweiss: As is it still looks like there's compilation breakage here -- this commit adds the third argument to the `UpdateBMCFirmware()` call, but that parameter isn't added to the function until 68546ed177468ab67569a65ad9a101e173f78c5f?

joelrebel: edited 808952d as e0b562ef6 and moved examples into fba84da

zevweiss (approved):

PR #209, opened 2021-03-17
Update Supermicro X11 PXE boot command

abdrabo:
The difference between the 2 commands is `chassis bootdev pxe` vs
`chassis bootdev pxe options=efiboot`. The one with options=efiboot does
not really work (if the server has OS installed on disk, it will boot
it).

Tested on Supermicro X11DSC+ with BIOS 3.1 and firmware 1.68.

### DIFF
--- a/providers/supermicro/supermicrox11/actions.go
+++ b/providers/supermicro/supermicrox11/actions.go
@@ -53,7 +53,7 @@ func (s *SupermicroX) PxeOnce() (status bool, err error) {
 	if err != nil {
 		return status, err
 	}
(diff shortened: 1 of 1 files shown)


jacobweinstock (approved):

abdrabo:
Looks similar to https://www.supermicro.com/support/faqs/faq.cfm?faq=33057

joelrebel:
I believe this needs to be reverted, since `chassis bootdev pxe` will not work for systems with UEFI enabled.

abdrabo:
OK I will revert it.

Issue #210, opened 2021-03-18
provide a way to tell the client which provider executed successfully


jacobweinstock:
while using `client.go` it's not known which provider successfully executed a call. It would be nice to have the ability to know which provider successfully executed a call.

jacobweinstock:
Functionality added in https://github.com/bmc-toolbox/bmclib/pull/212

PR #212, opened 2021-03-20
Identify the provider that executed successfully:

jacobweinstock:
Without breaking clients, this adds the capability of relaying to the end-user metadata about the method call, like the name of the provider that successfully executed. This gives info back to end-users to help if they need to use a certain provider for certain hardware, for other determinations that only happen on the end-user side, for troubleshooting, and removes some of the mystery around what happens when a method is called.

addresses https://github.com/bmc-toolbox/bmclib/issues/210

example usage in [`client_test.go`](https://github.com/bmc-toolbox/bmclib/pull/212/files#diff-4dc56e64f310cd8c12bb00f75c09e83f5d875563ea9901788e525487dc6788b3)

### DIFF
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -1,0 +1,9 @@
+package bmc
+
+// Metadata represents details about a bmc method
+type Metadata struct {
+	// SuccessfulProvider is the name of the provider that successfully executed
+	SuccessfulProvider string
+	// ProvidersAttempted is a slice of all providers that were attempt to execute
+	ProvidersAttempted []string
+}
--- a/bmc/bmc.go
+++ b/bmc/bmc.go
@@ -6,4 +6,8 @@ type Metadata struct {
 	SuccessfulProvider string
 	// ProvidersAttempted is a slice of all providers that were attempt to execute
 	ProvidersAttempted []string
+	// SuccessfulOpenConns is a slice of provider names that were opened successfully
+	SuccessfulOpenConns []string
+	// SuccessfulCloseConns is a slice of provider names that were closed successfully
+	SuccessfulCloseConns []string
 }
--- a/bmc/boot_device.go
+++ b/bmc/boot_device.go
@@ -21,7 +21,10 @@ type bootDeviceProviders struct {
 
 // SetBootDevice sets the boot device. Next boot only unless setPersistent=true
 // if a successfulProviderName is passed in, it will be updated to be the name of the provider that successfully executed
-func SetBootDevice(ctx context.Context, bootDevice string, setPersistent, efiBoot bool, b []bootDeviceProviders, successfulProviderName ...*string) (ok bool, err error) {
+func SetBootDevice(ctx context.Context, bootDevice string, setPersistent, efiBoot bool, b []bootDeviceProviders, metadata ...*Metadata) (ok bool, err error) {
+	if len(metadata) == 0 || metadata[0] == nil {
+		metadata = []*Metadata{&Metadata{}}
+	}
 Loop:
 	for _, elem := range b {
 		select {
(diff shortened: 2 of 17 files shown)


joelrebel:
The changes itself looks good to me and seem like a way to move ahead without breaking things for now,
although I've had a similar requirement with the [Open](https://github.com/bmc-toolbox/bmclib/blob/master/bmc/connection.go#L22) method.
 
As a change for later, I'm pondering is its worth that the Interface methods, 
all return a (Result, error) object,  where `Result` includes any information returned by the method,

```
type Result struct {
  SuccessfulProviderName string
  ProvidersAttempted []string
 ....
  Data interface{}
}
```

jacobweinstock:
> The changes itself looks good to me and seem like a way to move ahead without breaking things for now,
> although I've had a similar requirement with the [Open](https://github.com/bmc-toolbox/bmclib/blob/master/bmc/connection.go#L22) method.
> 
> As a change for later, I'm pondering is its worth that the Interface methods,
> all return a (Result, error) object, where `Result` includes any information returned by the method,
> 
> ```
> type Result struct {
>   SuccessfulProviderName string
>   ProvidersAttempted []string
>  ....
>   Data interface{}
> }
> ```

The open/close methods are different from all others because they try to open/close all providers. all other methods will stop once a successful method is executed. this PR https://github.com/bmc-toolbox/bmclib/pull/215 will remove any provider that doesn't successfully open so a user could inspect the returned drivers for which ones are there.

the challenge with "_all the Interface methods, all return_" this Result struct is that interface methods know nothing of each other. a power state get call from the ipmitool provider doesn't know anything about `ProvidersAttempted`. Also, returning something like `Result.Data` as an `interface{}` type means that methods could put different types in there and share this field, but at the sacrifice of the user experience. With this `Result.Data`, the user would then need to convert/type assert to the "correct" value. I don't think we should downgrade the user experience to make implementing providers easier.

An interesting idea is to possibly update the method helper functions like https://github.com/bmc-toolbox/bmclib/pull/212/files#diff-81e3810067afb632195aacdb034452b26f512d824218e33fc1bcc988a987494eR41 to take something like `successfulProviderName ...*Metadata` instead of `successfulProviderName ...*string`.  then we could put additional things in there. or we could refactor these method helper functions to also return some kind of a `Metadata` struct. I'm not seeing the value in changing the interface methods themselves though.

jacobweinstock:
@joelrebel , I've updated to use a metadata struct instead of just a string. Should allow us to get some of the info you suggested.

joelrebel:
> > The changes itself looks good to me and seem like a way to move ahead without breaking things for now,
> > although I've had a similar requirement with the [Open](https://github.com/bmc-toolbox/bmclib/blob/master/bmc/connection.go#L22) method.
> > As a change for later, I'm pondering is its worth that the Interface methods,
> > all return a (Result, error) object, where `Result` includes any information returned by the method,
> > ```
> > type Result struct {
> >   SuccessfulProviderName string
> >   ProvidersAttempted []string
> >  ....
> >   Data interface{}
> > }
> > ```
> 
> The open/close methods are different from all others because they try to open/close all providers. all other methods will stop once a successful method is executed. this PR #215 will remove any provider that doesn't successfully open so a user could inspect the returned drivers for which ones are there.
> 
> the challenge with "_all the Interface methods, all return_" this Result struct is that interface methods know nothing of each other. a power state get call from the ipmitool provider doesn't know anything about `ProvidersAttempted`. Also, returning something like `Result.Data` as an `interface{}` type means that methods could put different types in there and share this field, but at the sacrifice of the user experience. With this `Result.Data`, the user would then need to convert/type assert to the "correct" value. I don't think we should downgrade the user experience to make implementing providers easier.
> 
> An interesting idea is to possibly update the method helper functions like https://github.com/bmc-toolbox/bmclib/pull/212/files#diff-81e3810067afb632195aacdb034452b26f512d824218e33fc1bcc988a987494eR41 to take something like `successfulProviderName ...*Metadata` instead of `successfulProviderName ...*string`. then we could put additional things in there. or we could refactor these method helper functions to also return some kind of a `Metadata` struct. I'm not seeing the value in changing the interface methods themselves though.

Yep, agree there - it would not be to leave the user to deal with the result `interface{}, the `Metadata` approach is fine with me

joelrebel (approved):

PR #213, opened 2021-03-20
Add pure go ipmi provider:

jacobweinstock:
Add another provider. This provider has no external binary dependencies like the ipmitool provider does.


jacobweinstock:
requires https://github.com/bmc-toolbox/bmclib/pull/215 first

providers/goipmi/goipmi.go
```diff
@@ -0,0 +1,156 @@
...
+	return true, nil
+}
+
+func (c *Conn) status(ctx context.Context) (string, error) {
+	result := "off"
+	status, err := c.conn.GetChassisStatus(ctx)
+	if err != nil {
+		return "", err
+	}
+	if status.PoweredOn {
```
joelrebel, line 131: I think this is OK for now, although we may not want to hide the [other possible chassis states](https://github.com/gebn/bmc/blob/a5643220ab2ab78d974078c172817a1d4e92b7e4/pkg/ipmi/get_chassis_status.go#L220) from the end user?

providers/goipmi/goipmi.go
```diff
@@ -0,0 +1,156 @@
...
+	machine, err := bmc.Dial(ctx, c.Host)
+	if err != nil {
+		return err
+	}
+	c.transport = machine
+
+	sess, err := machine.NewSession(ctx, &bmc.SessionOpts{
+		Username:          c.User,
+		Password:          []byte(c.Pass),
+		MaxPrivilegeLevel: ipmi.PrivilegeLevelOperator,
```
joelrebel, line 50: Any reason we're going with the Operator priviledge here instead of [Administrator](https://github.com/gebn/bmc/blob/master/pkg/ipmi/privilege_level.go#L34)

joelrebel (changes requested):

providers/goipmi/goipmi.go
jacobweinstock: ah good catch, will update

providers/goipmi/goipmi.go
jacobweinstock: hmmm....im not seeing any other options besides [powered on or not](https://github.com/gebn/bmc/blob/a5643220ab2ab78d974078c172817a1d4e92b7e4/pkg/ipmi/get_chassis_status.go#L121).
https://github.com/gebn/bmc/blob/a5643220ab2ab78d974078c172817a1d4e92b7e4/cmd/describe/main.go#L298
https://github.com/gebn/bmc/blob/a5643220ab2ab78d974078c172817a1d4e92b7e4/pkg/ipmi/get_chassis_status.go#L96

client.go
```diff
@@ -75,6 +76,9 @@ func (c *Client) registerProviders() {
 	driverAsrockrack, _ := asrockrack.New(c.Auth.Host, c.Auth.User, c.Auth.Pass, c.Logger)
 	c.Registry.Register(asrockrack.ProviderName, asrockrack.ProviderProtocol, asrockrack.Features, nil, driverAsrockrack)
 
+	// register goipmi provider
```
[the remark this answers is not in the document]
jacobweinstock: _"does this mean we'd have two IPMI providers attempting a connection with Open"_ - yes, that is the case.
_"assuming the first IPMI provider fails ?"_ - no, the open method will try to open all providers specified. The providers can be filtered before an open though.

providers/goipmi/goipmi.go
jacobweinstock: updated and pushed

joelrebel (approved):

PR #223, opened 2021-04-08
Bug! We have to make sure that the chassis is on before power cycling!

omar-m-othman:
Without this change, we get "Set Chassis Power Control to Cycle failed: Command not supported in present state" if the chassis is turned off.

### DIFF
--- a/internal/ipmi/ipmi.go
+++ b/internal/ipmi/ipmi.go
@@ -69,6 +69,35 @@ func (i *Ipmi) PowerCycle(ctx context.Context) (status bool, err error) {
 	return false, fmt.Errorf("%v: %v", err, output)
 }
 
+// ForceRestart does the chassis power cycle even if the chassis is turned off.
+// From the RedFish spec (https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2018.1.pdf):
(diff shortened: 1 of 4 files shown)


joelrebel:
@jacobweinstock you might want to review this one

jacobweinstock:
I'm in favor of following the `ipmitool` definition of power cycle and redefining it to power a machine on if it is off.
_"cycle
Provides a power off interval of at least 1 second. No action should occur if chassis power is in S4/S5 state, but it is recommended to check power state first and only issue a power cycle command if the system power is on or in lower sleep state than S4/S5."_

jacobweinstock:
Hey @omar-m-othman , how do feel about this? want to discuss further?

omar-m-othman:
> Hey @omar-m-othman , how do feel about this? want to discuss further?

Sorry man! My notifications are f$%#ed up. I've just seen this now!

omar-m-othman:
> Hey @omar-m-othman , how do feel about this? want to discuss further?

I guess no. If you are sure that what we are doing matches the definition, then you are definitely right. Feel free to close the merge request.

nnuss (approved):

Issue #204, opened 2021-02-18
Add a golangci lint yaml


joelrebel:
.. and fix up resulting lint errors

PR #224, opened 2021-04-23
Fix bug in provider ipmitool:

jacobweinstock:
Power off in ipmitool was duplicating the internal ipmi
package functionality in PowerOff but doing it poorly.
Specifically not checking for an error when calling `IsOn`.

Also, the internal ipmi package `PowerOn` has been updated.
It now follows how ipmitool behaves when calling `chassis power on`.
ipmitool will return an exit code of 0 if calling `chassis power on`
against a machine in a powered-on state.

### DIFF
--- a/internal/ipmi/ipmi.go
+++ b/internal/ipmi/ipmi.go
@@ -112,11 +112,10 @@ func (i *Ipmi) PowerResetBmc(ctx context.Context, resetType string) (ok bool, er
 func (i *Ipmi) PowerOn(ctx context.Context) (status bool, err error) {
 	s, err := i.IsOn(ctx)
 	if err != nil {
-		return false, err
+		return false, errors.Wrap(err, "error checking power state")
 	}
-
 	if s {
-		return false, fmt.Errorf("server is already on")
+		return true, nil
 	}
 
 	output, err := i.run(ctx, []string{"chassis", "power", "on"})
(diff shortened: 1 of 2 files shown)


internal/ipmi/ipmi.go
```diff
@@ -112,11 +112,10 @@ func (i *Ipmi) PowerResetBmc(ctx context.Context, resetType string) (ok bool, er
 func (i *Ipmi) PowerOn(ctx context.Context) (status bool, err error) {
 	s, err := i.IsOn(ctx)
 	if err != nil {
-		return false, err
+		return false, errors.Wrap(err, "error checking power state")
```
joelrebel, line 115: I'd suggest the dynamic error here be defined as a type `ErrCheckPowerState`

joelrebel (approved):

internal/ipmi/ipmi.go
jacobweinstock: sorry, I'm not familiar enough with the design philosophy for custom errors in bmclib to feel comfortable adding this. Also since there are no other custom errors in this package already, this would be the first and that adds to the not comfortable feeling. Maybe another PR maybe for this?

joelrebel: Yep, can be done in a separate PR.
Since https://github.com/bmc-toolbox/bmclib/issues/204 updates the linter config which will catch these dynamic errors - I suggest we start defining newer errors at the top of any file they are being declared in, it'll just take off some linting work later.

PR #226, opened 2021-04-24
Update metadata user experience:

jacobweinstock:
Current:
```go
state, err := client.GetPowerState(ctx, &metadata)
if err != nil {
	panic(err)
}
fmt.Println(state)
fmt.Printf("metadata:  %+v\n", metadata)
```
Proposed:
```go
state, err := client.GetPowerState(ctx)
if err != nil {
	panic(err)
}
fmt.Println(state)
fmt.Printf("metadata: %+v\n", client.GetMetadata())
```
After using the metadata for a little, I don't like the
experience. Having to pass in metadata was not easily
understood and was trying to be too clever. This update
will make metadata available for all methods. the user just
has to call GetMetadata() to retrieve it.

Public API is affected as the client will no longer be able
to pass in the &metadata var. Seeing as we are not v1,
I feel comfortable proposing this change.

Signed-off-by: Jacob Weinstock <jakobweinstock@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ package bmclib
 import (
 	"context"
 	"io"
+	"sync"
 
 	"github.com/bmc-toolbox/bmclib/bmc"
 	"github.com/bmc-toolbox/bmclib/logging"
@@ -22,6 +23,7 @@ type Client struct {
 	Logger   logr.Logger
 	Registry *registrar.Registry
 	metadata *bmc.Metadata
+	mdLock   *sync.Mutex
 }
 
 // Auth details for connecting to a BMC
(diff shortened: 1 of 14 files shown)


client.go
```diff
@@ -21,6 +21,7 @@ type Client struct {
 	Auth     Auth
 	Logger   logr.Logger
 	Registry *registrar.Registry
+	metadata *bmc.Metadata
```
joelrebel, line 24: This approach is cleaner, and more user friendly than the previous, 
Although, since `Client.metadata` isn't under any mutex, the caller needs to be careful with interpreting the data held. I don't think this is a blocker to merge though, and given this object is now available it can be changed internally without breaking the API.

joelrebel (approved):

client.go
jacobweinstock: I can update with a mutex on writes. I only think this matters if users are making calls to multiple *Client.X functions/methods across goroutines.

PR #227, opened 2021-04-27
Update Client.Open method:

jacobweinstock:
The UX feels unnecessarily verbose/explicit. Most other
libraries that have open/close methods just return an error.
These libraries dont return multiple implementation but with
the ability to get metadata for each Client.X call I dont
think we need to worry about making it that explicit that
we've modified the providers. Open for discussion here, but
putting this in as a PR regardless.

Signed-off-by: Jacob Weinstock <jakobweinstock@gmail.com>


joelrebel (approved):
I'm in favour of this change, better now than later

PR #238, opened 2021-07-21
[WIP] Sensors data

joelrebel:
Adds interface and methods for Power draw, temperature, fan speeds and chassis health.

### DIFF
--- a/providers/redfish/redfish.go
+++ b/providers/redfish/redfish.go
@@ -28,7 +28,7 @@ var (
 	Features = registrar.Features{
 		providers.FeaturePowerSet,
 		providers.FeaturePowerState,
-		providers.FeaturePowersensors,
+		providers.FeaturePowerSensors,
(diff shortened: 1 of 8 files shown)


providers/redfish/redfish.go
```diff
@@ -26,17 +26,35 @@ var (
 	Features = registrar.Features{
 		providers.FeaturePowerSet,
 		providers.FeaturePowerState,
+		providers.FeaturePowersensors,
```
dustinmiller, line 29: might do `providers.FeaturePowerSensors,` instead

dustinmiller (dismissed):
lgtm

providers/redfish/redfish.go
joelrebel: thanks, fixed in 3c3eec6

joelrebel:
Old PR, closing

PR #239, opened 2021-10-08
Omar m othman/miscellaneous bug fixes

omar-m-othman:
All these changes have been tested thoroughly on thousands of servers using bmcbutler.


joelrebel:
> re, sure. I just happened to work for so long on this that I ended up mixing things up a bit. But if you prefer that I don't submit this one as-is, just let me know and I'll close this one and submit a few new ones asking @nnuss to approve immediately.

lets proceed with this PR as is, I'll add in a PR description template and so we can follow the points I mentioned above going ahead.

PR #246, opened 2021-11-23
Fix panic for nil metadata:

jacobweinstock:
This fixes a poor user experience of panicking on calls to metadata if the metadata is nil.


nnuss (approved):
Thanks, that is kinder

PR #258, opened 2022-02-01
Remove bou.ke/monkey

jacobweinstock:
* Checklist

- [ ] Tests added

The license for `bou.ke/monkey` prohibits the use of the library for any reason.
https://github.com/bouk/monkey/blob/master/LICENSE.md


joelrebel (approved):
Thanks! 
Since this reduces the test coverage of the package, we need to figure out another means of testing these methods going ahead.

joelrebel:
This could be an option for integration tests https://pkg.go.dev/github.com/rmxymh/infra-ecosphere#section-readme

PR #259, opened 2022-02-17
Don't error on ipmitool power off when machine is already off:

jacobweinstock:
The current behavior in `internal/ipmitool` is to error if a request for a power-off is called against an already powered-off machine. I proposed we remove this behavior and if a request for a power-off (in `internal/ipmitool`) is called against an already powered-off machine we return success. This also mirrors `ipmitool` itself. It does not error out when an already-off machine calls power off against it.

* Checklist

- [x] Tests added

### DIFF
--- a/internal/ipmi/ipmi.go
+++ b/internal/ipmi/ipmi.go
@@ -173,15 +173,6 @@ func (i *Ipmi) PowerOnForce(ctx context.Context) (status bool, err error) {
 
 // PowerOff power off the machine via bmc
 func (i *Ipmi) PowerOff(ctx context.Context) (status bool, err error) {
-	s, err := i.IsOn(ctx)
-	if err != nil {
-		return false, err
-	}
-
-	if !s {
-		return false, fmt.Errorf("server is already off")
-	}
-
 	output, err := i.run(ctx, []string{"chassis", "power", "off"})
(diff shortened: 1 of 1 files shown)


nshalman:
Is there a way to bubble up the fact that the machine was already off without having it be an error? Or should that happen at a higher level (software could check power status before issuing the poweroff)?

jacobweinstock:
> Is there a way to bubble up the fact that the machine was already off without having it be an error? Or should that happen at a higher level (software could check power status before issuing the poweroff)?

good question. The current function signature of power off doesn't make this possible. `PowerOff(ctx context.Context) (status bool, err error)`. I think a call to power state would probably be ideal.

nnuss:
:+1: from me.
I'm generally in favor of the functions reporting whether the device reached the requested state ( just as in #223 about power cycling ).

joelrebel:
In general, I'd suggest methods in bmclib perform just action requested and then its upto the caller to verify if the action was successful, by invoking another bmclib method.

In some cases a clear status is returned by the action itself, see for example https://github.com/bmc-toolbox/bmclib/blob/master/providers/dell/idrac8/actions.go#L39

joelrebel (approved):

PR #263, opened 2022-04-21
Add initial secure TLS options

micahhausler:
Signed-off-by: Micah Hausler <mhausler@amazon.com>

* Checklist
  - [x] Tests added

This PR adds secure TLS options for consumers of different providers. ~This does NOT yet allow the default bmclib client to enable TLS, as most of the client methods (`Open()`, `CreateUser()`, etc) and their corresponding interface implementations in `/bmc` (`UserCreator`, `UserReader`, etc) all instantiate connections without options.~

This depends on #265, and I'll rebase after that merges

  - asrock
  - idrac8
  - idrac9
  - m1000e
  - c7000
  - iLo
  - redfish
  - supermicro
  - supermicroX11

N/A

N/A

N/A

```
Add SecureTLS options for HTTP connections
```


examples/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
```
joelrebel, line 3: could this be moved under `/examples/pre.v1`

joelrebel (approved):
LGTM.

Thanks for your work here in fixing up all the providers to support connection options!  Could you also include an example for setting connection options for the "v1" bmclib client under `examples/v1/`.

PR #271, opened 2022-04-27
providers/ipmitool: Open() should attempt to connect

joelrebel:

This keeps the `Open` behaviour consistent with the other providers,
such that the provider is excluded if Open() fails.

### Checklist
- [ ] Tests added


micahhausler (approved):
lgtm

PR #274, opened 2022-05-09
errors: less verbose compatibility check errors

joelrebel:

compatibility check logs to log only when debug/trace logging is
enabled, a lot of compatibility checks are performed for each
connection.

### DIFF
--- a/providers/ipmitool/ipmitool.go
+++ b/providers/ipmitool/ipmitool.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"strings"
 
+	bmclibErrs "github.com/bmc-toolbox/bmclib/errors"
+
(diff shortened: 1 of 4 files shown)


providers/ipmitool/ipmitool.go
```diff
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"strings"
 
+	bmclibErrs "github.com/bmc-toolbox/bmclib/errors"
+
```
micahhausler, line 9: nit: why separate this on its own line?

micahhausler (approved):
LGTM

providers/ipmitool/ipmitool.go
joelrebel: was a spurious newline, purged.

PR #282, opened 2022-07-28
redfish/dell tasks: register task names for NIC, Drive, controllers

joelrebel:

This registers a few more Dell component update task names, these task names are used to identify firmware updates in progress.

- Dell


diogomatsubara (approved):
LGTM

PR #285, opened 2022-09-01
Initial HP SNMP support

L1Cafe:
### Checklist
- [ ] Tests added

HP

-

iLO4 and 5

-

```
Added initial SNMP support to enable and disable SNMP.
New configuration setting added to indicate enabling or disabling SNMP. Sample config available on [bmcbutler](https://github.com/bmc-toolbox/bmcbutler/blob/2073ba74749452ffa50e607fa193713541cb9ab1/samples/cfg/configuration.yml)
```

### DIFF
--- a/providers/hp/ilo/helpers.go
+++ b/providers/hp/ilo/helpers.go
@@ -100,6 +100,8 @@ func (i *Ilo) cmpAccessSettings(cfg *cfgresources.Network) (AccessSettings, bool
 		serialEnable = 2
 	}
 
+	// SNMP status is in cfg.SNMPEnable as a boolean
+
 	currentConfig, err := i.queryAccessSettings()
 	if err != nil {
 		return AccessSettings{}, false, err
@@ -135,6 +137,18 @@ func (i *Ilo) cmpAccessSettings(cfg *cfgresources.Network) (AccessSettings, bool
 		if currentConfig.VirtualMediaPort != cfg.KVMMediaPort {
 			return false
 		}
+		// Comparing SNMP settings for iLO 4 and iLO5
+		switch i.HardwareType() {
+		case "ilo4":
+			if (cfg.SNMPEnable && *currentConfig.SNMPSettings.SnmpExternalDisableIlo4 == 1) || (!cfg.SNMPEnable && *currentConfig.SNMPSettings.SnmpExternalDisableIlo4 == 0) {
+				return false
+			}
+		case "ilo5":
+			if (cfg.SNMPEnable && *currentConfig.SNMPSettings.SnmpExternalEnabledIlo5 == 0) ||
+				(!cfg.SNMPEnable && *currentConfig.SNMPSettings.SnmpExternalEnabledIlo5 == 1) {
+				return false
+			}
+		}
 
 		return true
 	}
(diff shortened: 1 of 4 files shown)


providers/hp/ilo/helpers.go
```diff
@@ -146,6 +160,28 @@ func (i *Ilo) cmpAccessSettings(cfg *cfgresources.Network) (AccessSettings, bool
 	currentConfig.IpmiPort = cfg.IpmiPort
 	currentConfig.SSHStatus = sshEnable
 	currentConfig.SSHPort = cfg.SSHPort
+	snmpSettings := new(SNMPSettings)
+	snmpSettings.SnmpPort = 161 // TODO: Change this to something user-configurable
+	snmpSettings.TrapPort = 162 // TODO: Change this to something user-configurable
+	switch i.HardwareType() {
```
joelrebel, line 166: same as my previous suggestion 
- implement a method to return the int pointer with the value set
- use the ilo4/ilo5 [constants](https://github.com/bmc-toolbox/bmclib/blob/58619e55ffc178bd531f9cda88fd9a328edf05dc/providers/hp/ilo/ilo.go#L33) in the case conditions

joelrebel (commented):
Thanks for this! left a few suggestions.

providers/hp/ilo/helpers.go
```diff
@@ -135,6 +137,18 @@ func (i *Ilo) cmpAccessSettings(cfg *cfgresources.Network) (AccessSettings, bool
 		if currentConfig.VirtualMediaPort != cfg.KVMMediaPort {
 			return false
 		}
+		// Comparing SNMP settings for iLO 4 and iLO5
+		switch i.HardwareType() {
```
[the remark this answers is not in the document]
L1Cafe: I have applied the constants in both cases.

Sorry, I don't understand what you mean with "move this the blocks under the case conditions into its own function". 

This section compares SNMP settings between the declared config and the config present on the BMC. Would you like me to make something like

`compareSNMPIlo4(...)` and `compareSNMPIlo5(...)` to avoid those long lines?

providers/hp/ilo/helpers.go
L1Cafe: I don't understand "return the int pointer with the value set".

Do you want an auxiliary function that creates the int pointer depending on hardware type and config, and just do something like

`currentConfig.SNMPSettings = setSnmpSettings(hardwareType, currentConfig)` ?

providers/hp/ilo/helpers.go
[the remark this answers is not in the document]
joelrebel: > I have applied the constants in both cases.
here I mean it could be `case Ilo4:` instead of `case "Ilo4"`.

> `compareSNMPIlo4(...)` and `compareSNMPIlo5(...)` to avoid those long lines?
yep 👍

providers/hp/ilo/helpers.go
joelrebel: yep 👍

L1Cafe:
@joelrebel please check if that is better

L1Cafe:
Hey @joelrebel I have the iDRAC8/9 SNMP implementation almost ready. Should I push to this repo to add the changes here or should we merge the code first?

joelrebel:
@L1Cafe a separate branch and PR would be ideal. 

You can go ahead and merge this in.

L1Cafe:
@joelrebel I don't think I can merge this on my own, it says "Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests."

Am I missing something?

joelrebel:
Should be fixed now, sent an invite for write access.

PR #289, opened 2022-10-17
Properly Handle Cases where the user Exists

jibsonline:
### Checklist
- [ NA] Tests added
- [ Y ] Similar commits squashed

HP
C7000
N/A
HP C7000
Currently we check for 400 status code while adding user to see if the user already exists.
But again we return error for the same status code(!200)

Changed that logic to proper;y check for existing users and not to throw error in such cases

### DIFF
--- a/providers/hp/c7000/configure.go
+++ b/providers/hp/c7000/configure.go
@@ -4,6 +4,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/bmc-toolbox/bmclib/cfgresources"
 	"github.com/bmc-toolbox/bmclib/devices"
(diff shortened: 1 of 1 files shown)


L1Cafe (approved):

providers/hp/c7000/configure.go
```diff
@@ -542,13 +543,18 @@ func (c *C7000) User(users []*cfgresources.User) (err error) {
...
 		payload := AddUser{Username: username, Password: password}
-		statusCode, _, err := c.postXML(payload)
+		statusCode, body, err := c.postXML(payload)
 		if err != nil {
 			allErrors += fmt.Sprintf("POST request to add user %s failed with error: %s", cfg.Name, err.Error())
 			continue
 		}
 
-		if statusCode == 400 {
+		userExists := false
```
joelrebel, line 552: by default a bool variable is false, this can just be `var userExists bool`

providers/hp/c7000/configure.go
```diff
@@ -575,8 +581,7 @@ func (c *C7000) User(users []*cfgresources.User) (err error) {
 				continue
 			}
 		}
-
-		if statusCode != 200 {
+		if statusCode != 200 && !userExists {
```
joelrebel, line 584: Its ideal to use`http.StatusOK` const instead of `200` for status code comparison.
https://go.dev/src/net/http/status.go

joelrebel (commented):
Added few nit pick comments, besides that LGTM.

providers/hp/c7000/configure.go
jibsonline: Done

L1Cafe (approved):

joelrebel (approved):
Thanks!

jibsonline:
@joelrebel I dont have write access to merge the the PR

PR #291, opened 2022-11-07
Redfish session fixes

joelrebel:

 - Fixes various superfluous nil error checks and remove calls to deprecated ioutil package.
 - Adds a check to ensure a Redfish session is active, this prevents the clients from seeing panics (see paste below)
   when a redfish session was closed or was not previously created.
 - A few inventory fixes.

Related: https://github.com/stmcginnis/gofish/pull/214

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x17b78fc]

goroutine 30 [running]:
github.com/bmc-toolbox/bmclib/v2/providers/redfish.(*Conn).status(0x17cb315?, {0xc0002f6470?, 0x0?})
        /go/pkg/mod/github.com/bmc-toolbox/bmclib/v2@v2.0.0/providers/redfish/redfish.go:236 +0x1c
github.com/bmc-toolbox/bmclib/v2/providers/redfish.(*Conn).PowerStateGet(0x18768c0?, {0x1b4ac18?, 0xc000433f80?})
        /go/pkg/mod/github.com/bmc-toolbox/bmclib/v2@v2.0.0/providers/redfish/redfish.go:178 +0x25
github.com/bmc-toolbox/bmclib/v2/bmc.getPowerState({0x1b4ac18, 0xc000433f80}, {0xc0003fa990?, 0x0?, 0x0?})
        /go/pkg/mod/github.com/bmc-toolbox/bmclib/v2@v2.0.0/bmc/power.go:102 +0x28d
github.com/bmc-toolbox/bmclib/v2/bmc.GetPowerStateFromInterfaces({0x1b4ac18?, 0xc000433f80?}, {0xc0002f6460?, 0x1?, 0x1?})
        /go/pkg/mod/github.com/bmc-toolbox/bmclib/v2@v2.0.0/bmc/power.go:131 +0x1c5
github.com/bmc-toolbox/bmclib/v2.(*Client).GetPowerState(0xc000196e10, {0x1b4ac18, 0xc000433f80})
        /go/pkg/mod/github.com/bmc-toolbox/bmclib/v2@v2.0.0/client.go:178 +0x1e7
```

### DIFF
--- a/providers/redfish/redfish.go
+++ b/providers/redfish/redfish.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	bmclibErrs "github.com/bmc-toolbox/bmclib/v2/errors"
-	"github.com/bmc-toolbox/bmclib/v2/internal"
 	"github.com/bmc-toolbox/bmclib/v2/internal/httpclient"
 	"github.com/bmc-toolbox/bmclib/v2/providers"
 	"github.com/go-logr/logr"
@@ -160,11 +159,6 @@ func (c *Conn) Compatible(ctx context.Context) bool {
 }
 
 func (c *Conn) sessionActive(ctx context.Context) error {
-	// skip session active checks for tests
-	if os.Getenv(internal.EnvBmclibTestActive) == "true" {
-		return nil
-	}
-
 	if c.conn == nil {
 		return bmclibErrs.ErrNotAuthenticated
 	}
--- a/providers/redfish/redfish.go
+++ b/providers/redfish/redfish.go
@@ -155,6 +155,11 @@ func (c *Conn) Compatible(ctx context.Context) bool {
 
 // BmcReset powercycles the BMC
 func (c *Conn) BmcReset(ctx context.Context, resetType string) (ok bool, err error) {
+	_, err = c.conn.GetSession()
+	if err != nil {
+		return false, errors.Wrap(bmclibErrs.ErrNotAuthenticated, err.Error())
+	}
+
 	managers, err := c.conn.Service.Managers()
 	if err != nil {
 		return false, err
@@ -172,11 +177,21 @@ func (c *Conn) BmcReset(ctx context.Context, resetType string) (ok bool, err err
 
 // PowerStateGet gets the power state of a BMC machine
 func (c *Conn) PowerStateGet(ctx context.Context) (state string, err error) {
+	_, err = c.conn.GetSession()
+	if err != nil {
+		return "", errors.Wrap(bmclibErrs.ErrNotAuthenticated, err.Error())
+	}
+
 	return c.status(ctx)
 }
 
 // PowerSet sets the power state of a BMC via redfish
 func (c *Conn) PowerSet(ctx context.Context, state string) (ok bool, err error) {
+	_, err = c.conn.GetSession()
+	if err != nil {
+		return false, errors.Wrap(bmclibErrs.ErrNotAuthenticated, err.Error())
+	}
+
 	switch strings.ToLower(state) {
 	case "on":
 		return c.on(ctx)
--- a/providers/redfish/redfish.go
+++ b/providers/redfish/redfish.go
@@ -122,7 +122,12 @@ func (c *Conn) Open(ctx context.Context) (err error) {
 
 // Close a connection to a BMC via redfish
 func (c *Conn) Close(ctx context.Context) error {
+	if c.conn == nil {
+		return nil
+	}
+
 	c.conn.Logout()
+
 	return nil
 }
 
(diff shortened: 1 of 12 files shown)


providers/redfish/redfish.go
```diff
@@ -156,8 +159,30 @@ func (c *Conn) Compatible(ctx context.Context) bool {
 	return err == nil
 }
 
+func (c *Conn) sessionActive(ctx context.Context) error {
+	// skip session active checks for tests
+	if os.Getenv(internal.EnvBmclibTestActive) == "true" {
```
jacobweinstock, line 164: any chance the `c.conn` can be replaced with a test implementation? This would remove the need for a global and test variable here. A global and test only variable living here means parallel tests are almost certainly out of the question.

joelrebel: Thanks for pointing that out, I've pushed a change that mocks the Redfish sessions endpoint instead.

chrisdoherty4: To a similar tune, if you replaced `c.conn` with an unexported interface you could swap it for a construct that deals with returning the error instead. I.e. this is a problem created by the fact `Close()` sets `c.conn` to `nil` and if it instead set `c.conn = sessionlessImplementation{}` we could avoid all the `c.sessionActive()` checks.

chrisdoherty4:
To a similar tune of https://github.com/bmc-toolbox/bmclib/pull/291#discussion_r1019482339.

If you replaced `c.conn` with an unexported interface you could swap it for a construct that deals with returning the error instead. I.e. this is a problem created by the fact `Close()` sets `c.conn` to `nil` and if it instead set `c.conn = sessionlessImplementation{}` we could avoid all the `c.sessionActive()` checks, I think.

joelrebel:
Since `c.conn` is a Gofish client from a dependency package - https://github.com/bmc-toolbox/bmclib/blob/8b19316c39fd2c2bc4b8fe448fd81fb2e1370da1/providers/redfish/redfish.go#L50

I'm not sure how we can go about implementing what you're suggesting, if you could elaborate a bit more that would help

To explain a bit more what I'm trying to fix here, 

The current bmclib methods call straight into the Gofish methods and or attempt to access Gofish client fields that may be nil which often results in panics,

```
func (c *Conn) reset(ctx context.Context) (ok bool, err error) {
	service := c.conn.Service <---  here
	ss, err := service.Systems()
	if err != nil {
		return false, err
	}
```

 The Gofish client or its fields may be happen to be nil because one of these cases,

  - The redfish session was not created in the first place - bmclib client never called `Open()`.
  - The redfish session was terminated by the BMC
  - The BMC has gone unresponsive. 

And so the main change here tries to guard the client against those cases by checking and returning a common error type,
https://github.com/bmc-toolbox/bmclib/blob/854838295f433e29aa57d8065bd069255c793b45/providers/redfish/redfish.go#L156

joelrebel:
> Side note, do we develop against mainline? I see this is into a v2 branch.

The intent is to have any new features and fixes go into the v2 branch which at some point will be the mainline.

The `master` branch is kept around since we have code bases that depend on it - for older hardware support and theres a few fixes going in there.

At the moment, I'm not seeing how we can transition to a single mainline completely while keeping the codebases separate. I'm open to any ideas or suggestions.

chrisdoherty4:
> if you could elaborate a bit more that would help

I think I see the confusion; I failed to understand the context some.

In short, I was trying to think of ways to wrap the underlying client (`gofish.APIClient`) such that _it_ handles the lack of a session without panicing rather than us instrumenting `sessionActive` in various methods. Doing so seems more maintainable as we don't need to remember to call `sessionActive`, it just works because of the wrapper.

For example

```go
type apiClientWrapper struct {
  *gofish.APIClient
}

func (c apiClientWrapper) Systems() ([]string, error) {
  if !c.isSessionActive() {
    return "", ErrNotAuthenticated
  }

  return c.APIClient.Systems()
}
```

```go

type apiClient interface {
  // Define all the behavior we need like Systems()
}

type Conn struct {
  conn apiClient // Use the interface instead of `*gofish.APIClient`.
}
```

_This is adapted on my previous comment having better understood the context._

--- 

> At the moment, I'm not seeing how we can transition to a single mainline completely while keeping the codebases separate. I'm open to any ideas or suggestions.

Probs best to chat off PR. I'm not familiar with how this has been released to date.

joelrebel:
Thanks, the`apiClientWrapper` with the embedded Gofish client is a better approach, I'll give that a try and update this PR.

joelrebel:
@chrisdoherty4 to do this correctly, I'd like to move out the redfish client wrapper from the package into `internal/redfish` and this way this abstraction would work as planned.

While I have some amount of this done - it seems to be a lot more work than I expected. I would like to take this up as a separate PR. And so for now, let's merge the current changes and I'll raise another for the redfish client abstraction.

chrisdoherty4 (approved):

jacobweinstock (approved):

PR #304, opened 2023-01-23
Rework intel amt provider to not panic:

jacobweinstock:
Panics were happening because of not handling return errors from `amt.NewClient`.

### Checklist
- [ ] Tests added

Intel AMT

Fix IntelAMT provider panics.
```
```


joelrebel (approved):
LGTM, thanks for the fix!

PR #305, opened 2023-01-27
Redfish versions param

joelrebel:

This enables the client to optionally define Redfish versions its not compatible
with, giving the user the option to ignore the Redfish endpoint on BMCs
for cases where the Redfish version is too old or is known to not have an expected functionality.

### Checklist
- [X] Tests added

```
- Adds the incompatible Redfish versions option parameter. 
```

### DIFF
--- a/internal/redfishwrapper/client.go
+++ b/internal/redfishwrapper/client.go
@@ -12,17 +12,19 @@ import (
 	"github.com/bmc-toolbox/bmclib/v2/internal/httpclient"
 	"github.com/pkg/errors"
 	"github.com/stmcginnis/gofish"
+	"golang.org/x/exp/slices"
 )
 
 // Client is a redfishwrapper client which wraps the gofish client.
 type Client struct {
-	host                 string
-	port                 string
-	user                 string
-	pass                 string
-	client               *gofish.APIClient
-	httpClient           *http.Client
-	httpClientSetupFuncs []func(*http.Client)
+	host                  string
+	port                  string
+	user                  string
+	pass                  string
+	versionsNotCompatible []string // a slice of redfish versions to ignore as incompatible
+	client                *gofish.APIClient
+	httpClient            *http.Client
+	httpClientSetupFuncs  []func(*http.Client)
 }
 
 // Option is a function applied to a *Conn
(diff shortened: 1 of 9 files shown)


internal/redfishwrapper/client.go
```diff
@@ -91,11 +104,8 @@ func (c *Client) Open(ctx context.Context) error {
 
 	var err error
 	c.client, err = gofish.ConnectContext(ctx, config)
-	if err != nil {
```
jacobweinstock, line 94: Was this intentional to remove the error check?

joelrebel:
> Thanks Joel. Curious, how is the current behavior between incompatible redfish versions and bmclib manifested? I'm assuming it's in some very inconsistent way so it's difficult to code-ify?

@jacobweinstock we have some BMCs where the Redfish user accounts are not connected to the actual BMC user accounts in any way - and the User create/update methods work, but not as expected. The vendor BMC Redfish is still "under development".

This change enables us to work around those BMCs by skipping the Redfish endpoint on them and trying the `vendorapi` endpoint instead. This would also help other cases where a firmware update through the Redfish endpoint may not work and so we'd bmclib to use the `vendorapi` instead.

internal/redfishwrapper/client.go
joelrebel: Yep it was intentional, the error check was a superfluous - since its not doing anything with the error there. This was caught by one of the semgrep rules I have setup https://github.com/dgryski/semgrep-go/blob/master/errnilcheck.yml

splaspood (approved):

jacobweinstock:
> > Thanks Joel. Curious, how is the current behavior between incompatible redfish versions and bmclib manifested? I'm assuming it's in some very inconsistent way so it's difficult to code-ify?
> 
> @jacobweinstock we have some BMCs where the Redfish user accounts are not connected to the actual BMC user accounts in any way - and the User create/update methods work, but not as expected. The vendor BMC Redfish is still "under development".
> 
> This change enables us to work around those BMCs by skipping the Redfish endpoint on them and trying the `vendorapi` endpoint instead. This would also help other cases where a firmware update through the Redfish endpoint may not work and so this change would get bmclib to use the `vendorapi` instead.

Thanks Joel, very helpful info to have in the open.

I have a concern that is change is very machine, vendor, etc specific and a holistically inflexible solution. For example, because specifying unsupported versions is a build time of the `Client`, it would seem that all functionality for unsupported versions would be skipped. So if i wanted to skip just user functionality and not power actions, this wouldnt be possible.

I would propose that we push this decision logic up to the client through filtering of the registered providers. this would allow runtime filtering that is more flexible. We (I) could introduce a new functionality in the registrar library for skipping. Then a user could construct a `Client` once and then query a BMC for its version and skip redfish for just user actions and then "unskip" after user actions. Thoughts?

joelrebel:
> I would propose that we push this decision logic up to the client through filtering of the registered providers. this would allow runtime filtering that is more flexible. We (I) could introduce a new functionality in the registrar library for skipping. Then a user could construct a Client once and then query a BMC for its version and skip redfish for just user actions and then "unskip" after user actions. Thoughts?

This runtime based filtering of the providers seems like an ideal option here, let me know what you're able to come up with and I'd be keen on testing it out. 

I still think it's worth having the option to filter out Redfish versions that the caller may know are incompatible - so we could have both these mechanisms, as long we document them and the related pros/cons?

jacobweinstock (approved):

joelrebel:
Added a few notes on the BMC connection filtering mechanisms so its a bit more clear on how they can be used https://github.com/bmc-toolbox/bmclib/pull/305/commits/2a1f02f77f65214c5a5238f08384dbb3fd1c4ecf

jacobweinstock (approved):

PR #307, opened 2023-01-31
Support redfish VirtualMedia

fintelia:

This PR adds support for setting redfish virtual media. It can be used, for instance, to specify the URL of an ISO image that the machine should boot from.

### Checklist
- [x] Tests added

Redfish. Tested on a Dell R640 server.

```
Added Redfish VirtualMedia support
```

### DIFF
--- a/bmc/virtual_media.go
+++ b/bmc/virtual_media.go
@@ -1,0 +1,70 @@
+package bmc
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+// VirtualMediaSetter controls the virtual media attached to a machine
+type VirtualMediaSetter interface {
+	SetVirtualMedia(ctx context.Context, kind string, mediaURL string) (ok bool, err error)
+}
+
+// VirtualMediaProviders is an internal struct to correlate an implementation/provider and its name
+type virtualMediaProviders struct {
+	name               string
+	virtualMediaSetter VirtualMediaSetter
+}
+
+// setVirtualMedia sets the virtual media.
+func setVirtualMedia(ctx context.Context, kind string, mediaURL string, b []virtualMediaProviders) (ok bool, metadata Metadata, err error) {
+	var metadataLocal Metadata
+Loop:
+	for _, elem := range b {
+		if elem.virtualMediaSetter == nil {
+			continue
+		}
+		select {
+		case <-ctx.Done():
+			err = multierror.Append(err, ctx.Err())
+			break Loop
+		default:
+			metadataLocal.ProvidersAttempted = append(metadataLocal.ProvidersAttempted, elem.name)
+			ok, setErr := elem.virtualMediaSetter.SetVirtualMedia(ctx, kind, mediaURL)
+			if setErr != nil {
+				err = multierror.Append(err, errors.WithMessagef(setErr, "provider: %v", elem.name))
+				continue
+			}
+			if !ok {
+				err = multierror.Append(err, fmt.Errorf("provider: %v, failed to set virtual media", elem.name))
+				continue
+			}
+			metadataLocal.SuccessfulProvider = elem.name
+			return ok, metadataLocal, nil
+		}
+	}
+	return ok, metadataLocal, multierror.Append(err, errors.New("failed to set virtual media"))
+}
+
+// SetVirtualMediaFromInterfaces identifies implementations of the virtualMediaSetter interface and passes the found implementations to the setVirtualMedia() wrapper
+func SetVirtualMediaFromInterfaces(ctx context.Context, kind string, mediaURL string, generic []interface{}) (ok bool, metadata Metadata, err error) {
+	bdSetters := make([]virtualMediaProviders, 0)
+	for _, elem := range generic {
+		temp := virtualMediaProviders{name: getProviderName(elem)}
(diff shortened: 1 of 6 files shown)


internal/redfishwrapper/virtual_media.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	for _, manager := range managers {
+		virtualMedia, err := manager.VirtualMedia()
+		if err != nil {
+			return false, err
+		}
+		for _, media := range virtualMedia {
+			if media.Inserted {
+				err = media.EjectMedia()
```
joelrebel, line 39: This loop seems to eject all existing virtualMedia, is this the expected behaviour - should this only eject the mediaKind its attempting to set?

internal/redfishwrapper/virtual_media.go
```diff
@@ -0,0 +1,74 @@
...
+				if err != nil {
+					return false, err
+				}
+			}
+		}
+	}
+
+	if mediaUrl != "" {
```
joelrebel, line 47: Is `mediaUrl` a required parameter, if so, this method could return early if its not specified.

internal/redfishwrapper/virtual_media.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	rf "github.com/stmcginnis/gofish/redfish"
+)
+
+// Set the boot device for the system.
+func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaUrl string) (ok bool, err error) {
```
joelrebel, line 12: s/mediaUrl/mediaURL/

joelrebel (commented):
Thanks for work and contribution here, I've left a few suggestions.

bmc/virtual_media.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// VirtualMediaProviders is an internal struct to correlate an implementation/provider and its name
+type virtualMediaProviders struct {
+	name               string
+	virtualMediaSetter VirtualMediaSetter
+}
+
+// setVirtualMedia sets the virtual media.
+func setVirtualMedia(ctx context.Context, kind string, mediaUrl string, b []virtualMediaProviders) (ok bool, metadata Metadata, err error) {
```
joelrebel, line 23: can we have tests for the methods in this file, hoping at some point we could generate the boilerplate code here along with tests.

joelrebel (changes requested):

internal/redfishwrapper/virtual_media.go
fintelia: Happy to change it. I wasn't 100% sure whether BMCs allow multiple kinds of virtual media to be attached at once, and this approach guarantees there won't be.

internal/redfishwrapper/virtual_media.go
fintelia: Right now setting `mediaUrl = ""`  causes virtual media to be ejected. That seemed simpler than adding a whole separate method.

internal/redfishwrapper/virtual_media.go
joelrebel: Sure, we can go with this approach for now.

internal/redfishwrapper/virtual_media.go
joelrebel: oh right, now it makes sense - could you leave a comment on there unless I've missed it.

bmc/virtual_media.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return ok, metadataLocal, multierror.Append(err, errors.New("failed to set virtual media"))
+}
+
+// SetVirtualMediaFromInterfaces identifies implementations of the virtualMediaSetter interface and passes the found implementations to the setVirtualMedia() wrapper
+func SetVirtualMediaFromInterfaces(ctx context.Context, kind string, mediaURL string, generic []interface{}) (ok bool, metadata Metadata, err error) {
+	bdSetters := make([]virtualMediaProviders, 0)
+	for _, elem := range generic {
+		temp := virtualMediaProviders{name: getProviderName(elem)}
+		switch p := elem.(type) {
```
chrisdoherty4, line 57: @joelrebel are we using go 1.18+? If so this could be simplified using generics, I think.

client.go
```diff
@@ -240,6 +240,13 @@ func (c *Client) SetBootDevice(ctx context.Context, bootDevice string, setPersis
 	return ok, err
 }
 
+// SetVirtualMedia pass through to library functino
```
chrisdoherty4, line 243: From the client consumers perspective, what does calling this do? I think that'd be helpful to outline in the method comment.

bmc/virtual_media.go
joelrebel: yep we are on 1.18 I have to get round to giving generics a try for this use case.

fintelia: I just based this function on the other existing implementations. I'm not familiar with go's generics, so it might make more sense for someone else to convert all of them in a follow up PR?

chrisdoherty4: I'm not intimately familiar with the code base so I didn't realize there were a bunch of similar things. All good.

joelrebel (approved):
Thanks for the fixes!

PR #308, opened 2023-02-14
redfish: added support for adapters with multiple ports

EugenioSiciliano:
## What does this PR implement/change/remove?
Multiple ports in the same network adapter are now inventoried as different NICs.

PR #309, opened 2023-02-21
redfish: added support for adapters with multiple ports

EugenioSiciliano:
Reworked according to discussion in https://github.com/bmc-toolbox/bmclib/pull/308.

Multiple ports in the same network adapter are now inventoried as different NICPorts under the same NIC.
Added some network port fields.
Fixed network speed conversion.
Added unit tests for network port and ethernet interface processing.

```
Added support for NICs with multiple ports.
Fixed network speed conversion.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/bmc-toolbox/bmclib/v2
 go 1.18
 
 require (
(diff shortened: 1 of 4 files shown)


joelrebel (approved):
Thanks for the work here along with the tests!

PR #312, opened 2023-03-08
Update Intel AMT library:

jacobweinstock:
The previous Intel AMT libraries were forks and the hope was that I would be able to upstream the changes. The upstreams themselves were actually forks and the PRs I opened have been sitting for some time. Because of all of this, I have created a new repo that I'll be maintaining. The new repo handles context timeouts across the board.

### Checklist
- [ ] Tests added

```
```


providers/intelamt/intelamt.go
```diff
@@ -125,15 +103,16 @@ func (c *Conn) BootDeviceSet(ctx context.Context, bootDevice string, setPersiste
 
 // PowerStateGet gets the power state of a BMC machine
 func (c *Conn) PowerStateGet(ctx context.Context) (state string, err error) {
+	fmt.Println("PowerStateGet")
```
joelrebel, line 106: debugging print?

joelrebel (changes requested):
Thanks for the work here! apart from the debug print, it looks good to me.

providers/intelamt/intelamt.go
jacobweinstock: whoops, yeah will remove. thanks

joelrebel (approved):

PR #313, opened 2023-03-09
Add timeouts everywhere

jacobweinstock:
This expands the timeout for opening connections to all method/BMC interactions.

If a context with a timeout IS provided to a method call but no per provider timeout is defined, then we default to a timeout of the provided timeout duration divided by the number of providers/drivers in the Client.Registry.Drivers slice.

If a context with a timeout IS NOT provided to a method call and no per provider timeout is defined, then we use the defaultConnectTimeout of 30 seconds.

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -24,20 +24,20 @@ import (
 
 // Client for BMC interactions
 type Client struct {
-	Auth           Auth
-	connectTimeout time.Duration
-	Logger         logr.Logger
-	Registry       *registrar.Registry
-	metadata       *bmc.Metadata
-	mdLock         *sync.Mutex
+	Auth     Auth
+	Logger   logr.Logger
+	Registry *registrar.Registry
 
-	redfishVersionsNotCompatible []string
 	httpClient                   *http.Client
 	httpClientSetupFuncs         []func(*http.Client)
+	mdLock                       *sync.Mutex
+	metadata                     *bmc.Metadata
+	perProviderTimeout           func(context.Context) time.Duration
+	redfishVersionsNotCompatible []string
 }
 
-var (
-	// default connection open timeout
+const (
+	// default connection timeout
 	defaultConnectTimeout = 30 * time.Second
 )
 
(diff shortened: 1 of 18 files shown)


client.go
```diff
@@ -77,12 +77,16 @@ func WithHTTPClient(c *http.Client) Option {
...
 }
 
-// WithConnectTimeout sets the timeout when connecting to a BMC to create a new session.
-// This timeout value applies for each provider.
-// When not defined the default connection timeout applies.
-func WithConnectTimeout(t time.Duration) Option {
+// WithPerProviderTimeout sets the timeout when interacting with a BMC.
+// This timeout value is applied per provider.
+// When not defined and a context with a timeout is passed to a method, the default timeout
+// will be the context timeout duration divided by the number of providers in the registry,
```
joelrebel, line 83: I'm a bit divided (pun intended) on dividing the timeout between the providers since it can have a really crappy experience - if an action requires more time than the slice it gets, then to fix it, the user has to account for the time being sliced between providers and that a single provider could consume a lot of it. 

For now I guess this is fair solution to honor the context timeout value. 

Going ahead we could figure if the `OpenConnectionFromInterfaces` method can attempt to open connection through the providers concurrently, thus keeping the timeout for each provider separate. And then, if the provider exports the feature the user is invoking - say power cycle, then we don't cycle through all of the providers, except those that export the feature, if that makes sense? Or does the registry package currently do that?

joelrebel (commented):
Thanks for fixing all of these up to honor the timeout.

Can you update the README.md to mention the `WithPerProviderTimeout` is recommended for the best experience when dealing with multiple vendors, I can update the examples once this is merged.

client.go
jacobweinstock: I think we're in an interesting position with trades off either way. Without slicing the timeout, the first or any single provider could take the whole timeout and not leave anytime for any other providers to try. Not ideal user experience either.

If `context.WithTimeout` is provided I think we should honor that as the max timeout. 

I like your idea around concurrent attempts for `OpenConnectionFromInterfaces`.  

In regard to features, like `providers.FeaturePowerSet`, those are for user side filtering only. some thing like, `Client.Registry.Supports(providers.FeaturePowerSet)`.  In regard to only attempting providers that implement a feature, we do that but not with regard to features but with regard to interfaces implemented. So if a provider doesn't implement user creation, for example, we dont attempt it. see [here](https://github.com/bmc-toolbox/bmclib/blob/4d46ab246d57c50363a24b53634ee7e28a5b82cf/bmc/user.go#L76).

I will update the README.md with details about timeouts. Thanks for reviewing this!

joelrebel (approved):

PR #316, opened 2023-03-21
Redfish timeout update

### LINKED ISSUE #315: Redfish provider broken
The recent [timeout changes](https://github.com/bmc-toolbox/bmclib/commit/100c29496b91320b9f969a38a3a5f46ada4fcb4d#diff-32b61b6ce812301d1fec53a45d9303a679bb9289762bc0edd077c4cb5cb11867R48) in [OpenConnectionFromInterfaces](https://github.com/bmc-toolbox/bmclib/blob/main/bmc/connection.go#L31) breaks the Redfish provider.

This is because the context to the Redfish client is passed through bmclibs `Open()` method and the Redfish client holds the context in its [APIClient struct]( https://github.com/stmcginnis/gofish/blob/main/client.go#L38) the context passed is then canceled when `OpenConnectionFromInterfaces` returns, and so any further calls on the client return a context canceled error.

The option here seems to be that,
 - `OpenConnectionFromInterfaces` returns the context `cancel` funcs for successful providers, which are then kept in the bmclib client, and are then canceled when the caller invokes `Close()`.
 
@jacobweinstock let me know if you have any other/better ideas.

To reproduce the issue, test with this modified inventory example

```go
package main

import (
	"context"
	"crypto/x509"
	"encoding/json"
	"flag"
	"fmt"
	"log"
	"os"
	"strconv"
	"strings"
	"time"

	bmclib "github.com/bmc-toolbox/bmclib/v2"
	"github.com/bombsimon/logrusr/v2"
	"github.com/sirupsen/logrus"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
	defer cancel()

	user := flag.String("user", "", "Username to login with")
	pass := flag.String("password", "", "Username to login with")
	host := flag.String("host", "", "BMC hostname to connect to")
	port := flag.Int("port", 443, "BMC port to connect to")
	withSecureTLS := flag.Bool("secure-tls", false, "Enable secure TLS")
	certPoolFile := flag.String("cert-pool", "", "Path to an file containing x509 CAs. An empty string uses the system CAs. Only takes effect when --secure-tls=true")
	flag.Parse()

	l := logrus.New()
	l.Level = logrus.DebugLevel
	logger := logrusr.New(l)

	clientOpts := []bmclib.Option{bmclib.WithLogger(logger)}

	if *withSecureTLS {
		var pool *x509.CertPool
		if *certPoolFile != "" {
			pool = x509.NewCertPool()
			data, err := os.ReadFile(*certPoolFile)
			if err != nil {
				l.Fatal(err)
			}
			pool.AppendCertsFromPEM(data)
		}
		// a nil pool uses the system certs
		clientOpts = append(clientOpts, bmclib.WithSecureTLS(pool))
	}

	cl := bmclib.NewClient(*host, strconv.Itoa(*port), *user, *pass, clientOpts...)

	cl.Registry.Drivers = cl.Registry.Using("redfish")

	err := cl.Open(ctx)
	if err != nil {
		log.Fatal(err, "bmc login failed")
	}

	defer cl.Close(ctx)

	inv, err := cl.Inventory(ctx)
	if err != nil {
		l.Error(err)
	}

	b, err := json.MarshalIndent(inv, "", "  ")
	if err != nil {
		l.Error(err)
	}

	fmt.Println(string(b))
}

```

jacobweinstock:

Fix redfish provider's context timed out/canceled issue.

Fixes #315 

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ tags
 *.code-workspace
 # added by lint-install
 out/
+
+coverage.txt
--- a/bmc/connection.go
+++ b/bmc/connection.go
@@ -76,21 +76,14 @@ func closeConnection(ctx context.Context, c []connectionProviders) (metadata Met
 		if elem.closer == nil {
 			continue
 		}
-		select {
-		case <-ctx.Done():
-			err = multierror.Append(err, ctx.Err())
-
-			return metadata, err
-		default:
-			metadataLocal.ProvidersAttempted = append(metadataLocal.ProvidersAttempted, elem.name)
-			closeErr := elem.closer.Close(ctx)
-			if closeErr != nil {
-				err = multierror.Append(err, errors.WithMessagef(closeErr, "provider: %v", elem.name))
-				continue
-			}
-			connClosed = true
-			metadataLocal.SuccessfulCloseConns = append(metadataLocal.SuccessfulCloseConns, elem.name)
+		metadataLocal.ProvidersAttempted = append(metadataLocal.ProvidersAttempted, elem.name)
+		closeErr := elem.closer.Close(ctx)
+		if closeErr != nil {
+			err = multierror.Append(err, errors.WithMessagef(closeErr, "provider: %v", elem.name))
+			continue
 		}
+		connClosed = true
+		metadataLocal.SuccessfulCloseConns = append(metadataLocal.SuccessfulCloseConns, elem.name)
 	}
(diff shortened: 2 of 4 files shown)


joelrebel (approved):
thanks for this fix, looks good to me!

PR #318, opened 2023-04-15
Redfish nextboot update

jacobweinstock:
Setting the boot device on HP DL160 machines was failing. There were too many values in the payload that were being rejected. Only sending the BootSourceOverrideEnabled and BootSourceOverrideTarget got this call working on the HP.

This was working fine before this change against a Dell R240. With this change the Dell R240 was still working. Hopefully this is will enable support for more systems generally.

The efiboot setting was also causing the set boot device redfish call to fail on the HP machine. I didn't seem to have any affect on the Dell R240 machine so i removed it entirely.

Hey @joelrebel , If you're able, I'd love to have to test this against some other models. Thanks!

### Checklist
- [ ] Tests added

```
```


joelrebel:
yep, give me a few days I'll give it a spin.

joelrebel:
Looks good to me, tested on Dell R6515, R740, OpenBMC and few Supermicros.

joelrebel (approved):

PR #322, opened 2023-04-28
Refactor BMC provider opener to probe concurrently

chrisdoherty4:
## Summary

In the Tinkerbell project where we leverage bmclib in Cluster API contexts we've had reports of BMC interactions taking too long resulting in CAPI components rolling nodes it thinks are faulty. As part of the investigation we dived into the timeout capabilities available in bmclib when connections/interfaces are initial established/probed and found a user can either (1) specify a static timeout that's shared among all providers (IPMI, Redfish etc) or (2) specify a per provider timeout that will linearly grow the total timeout applied with the total providers attempted.

This change alters the algorithm to open connections to operate concurrently. By operating concurrently we stand a better chance of giving each provider the maximal timeout period to establish its connection and avoid the linear growth with per provider timeouts.

We haven't benchmarked this change and acknowledge systems that don't offload network IO or are single CPU won't necessarily benefit but given modern consumer hardware is generally multicore we anticipate it will bring a general improvement.


bmc/connection.go
```diff
@@ -28,67 +29,105 @@ type connectionProviders struct {
...
+	case <-ctx.Done():
+		err = multierror.Append(err, ctx.Err())
+		return nil, metadata, err
+	default:
+	}
+
+	// Create a context with the specified timeout. This is done for backward compatibility but
+	// we should consider removing the timeout parameter alltogether given the context will
+	// container the timeout.
+	ctx, cancel := context.WithTimeout(ctx, timeout)
```
jacobweinstock, line 46: this feels like it should be right next to `o.Open` and a new context created per provider. otherwise, each provider could affect the context of another.  This is one reason that the current implementation creates a new context the way it does.

bmc/connection.go
```diff
@@ -28,67 +29,105 @@ type connectionProviders struct {
...
 		default:
-			e := fmt.Sprintf("not a Opener implementation: %T", p)
-			err = multierror.Append(err, errors.New(e))
+			err = multierror.Append(err, fmt.Errorf("not a Opener implementation: %T", p))
+		}
+	}
+
+	// Launch the goroutine to close the results channel ensuring we can exit the for-range over
+	// the results channel below.
+	go func() { wg.Wait(); close(results) }()
```
jacobweinstock, line 90: just food for thought, if we dont use a `WaitGroup` but instead in the range over results we do a count and compare that to the `len(providers)` then we save a few allocations and use less memory. 

```bash
100	  10704031 ns/op	    2686 B/op	      46 allocs/op (with WaitGroup)
100	  10643783 ns/op	    2415 B/op	      44 allocs/op (without WaitGroup)
```

See the benchmark test i used here: https://github.com/jacobweinstock/bmclib/blob/13118c0efcf49ddbdb6e580675442fa0a79da095/bmc/connection_test.go#L63

jacobweinstock (changes requested):
Thanks for this @chrisdoherty4, you beat me to it! :) I was [almost ready](https://github.com/jacobweinstock/bmclib/blob/add-concurrency-to-open/bmc/connection.go) to open a PR. Overall looks good. I have one comment and one non-blocking idea.

bmc/connection.go
```diff
@@ -28,67 +29,105 @@ type connectionProviders struct {
...
+
+	// Gather and handle results from the opener goroutines.
+	for res := range results {
+		if res.Err == nil {
+			opened = append(opened, res.Opener)
+			metadata.SuccessfulOpenConns = append(metadata.SuccessfulOpenConns, res.ProviderName)
+		} else {
+			err = multierror.Append(err, errors.WithMessagef(res.Err, "provider: %v", res.ProviderName))
```
jacobweinstock, line 98: I think the `WithMessagef` is causing the test failure. You could probably just `err = multierror.Append(err, res.Err)`.

joelrebel:
Thanks for the efforts here! Could we also include a [check for Goroutine leaks](https://github.com/uber-go/goleak) in `TestOpenConnectionFromInterfaces` so its covered for now and any changes going ahead.

chrisdoherty4:
> I was [almost ready](https://github.com/jacobweinstock/bmclib/blob/add-concurrency-to-open/bmc/connection.go?rgh-link-date=2023-04-28T04%3A25%3A58Z) to open a PR.

Nice, I did this quite late so haven't verified everything. Lets compare the approaches and see what we prefer.

bmc/connection.go
chrisdoherty4: What effects do you think could propagate between them?

chrisdoherty4:
> Thanks for the efforts here! Could we also include a [check for Goroutine leaks](https://github.com/uber-go/goleak) in TestOpenConnectionFromInterfaces so its covered for now and any changes going ahead.

Overlooked this comment, will add it in.

chrisdoherty4:
@joelrebel The goleak tests should be in-place.

bmc/connection_test.go
```diff
@@ -66,6 +67,8 @@ func TestOpenConnectionFromInterfaces(t *testing.T) {
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
+			defer goleak.VerifyNone(t)
```
joelrebel, line 70: Based on the goleak docs, couldn't this just be defined at the top in the test function, instead of being part of the loop?

chrisdoherty4: It could be, but then how would you know which sub test resulted in the leak? It's akin to putting it in TestMain, we wouldn't know what test triggered the leak.

joelrebel (approved):
Seems to work just fine in my testing, thanks for the fixes!

bmc/connection.go
chrisdoherty4: @jacobweinstock

jacobweinstock (approved):

PR #326, opened 2023-04-30
Add per provider config:

jacobweinstock:
There exists some config that cannot be shared across all providers. For example, the port for ipmitool is generally 623 but is generally 16992 for intelAMT, cipher suites for ipmitool, or TLS certificates. Allowing per-provider config enables customizing this type of data.

This introduces a * *__breaking change__* * to `bmclib.NewClient`. It no longer takes in a port. The ports for each provider have a default defined and each one can be overridden with a functional option. `WithIpmitoolPort`, for example.

I think fundamentally the behavior from the consumer side doesn't change. Consumers will still use `With` functions to specify per provider config. This change just organizes it differently under the hood.  

### Checklist
- [ ] Tests added

```
```


client.go
```diff
@@ -90,22 +89,28 @@ func WithPerProviderTimeout(timeout time.Duration) Option {
...
-
 // NewClient returns a new Client struct
 func NewClient(host, port, user, pass string, opts ...Option) *Client {
 	var defaultClient = &Client{
-		Logger:                       logr.Discard(),
-		Registry:                     registrar.NewRegistry(),
-		redfishVersionsNotCompatible: []string{},
+		Logger:   logr.Discard(),
+		Registry: registrar.NewRegistry(),
+		providerConfig: providerConfig{
```
joelrebel, line 97: Is it possible to have the `providerConfig` (or have Options) type be exported as a struct/interface by each provider -  such provider specific configuration and http client checks can stay within the respective packages under `providers/{asrockrack,redfish..}/` ?

This would keep the client package free from provider specifics.

joelrebel (commented):
Thanks for the efforts here, left a question/comment inline..

client.go
jacobweinstock: Hey @joelrebel , thanks for looking at this. I've updated to have all providers define their own config struct and then client.go just collects them into a single struct. Is this more of what you had in mind? Also, this PR is still very much a WIP.

joelrebel: Will take a look thanks!, apologies I missed the WIP/draft label on the PR.

option.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+func WithAsrockrackPort(port string) Option {
+	return func(args *Client) {
+		args.providerConfig.asrock.Port = port
+	}
+}
+
+func WithGofishHTTPClient(httpClient *http.Client) Option {
```
joelrebel, line 85: As a library user with no clue what Gofish is, `Redfish` would be clearer in this case, since the user does not have to know what Gofish refers to. And while Gofish enables this library to connect to Redfish, I'm a bit conflicted with this change, but its not a blocker in anycase.

client.go
```diff
@@ -163,25 +120,41 @@ func (c *Client) defaultTimeout(ctx context.Context) time.Duration {
...
+	}
+	if driverIpmitool, err := ipmitool.New(c.Auth.Host, c.Auth.User, c.Auth.Pass, ipmiOpts...); err == nil {
+		c.Registry.Register(ipmitool.ProviderName, ipmitool.ProviderProtocol, ipmitool.Features, nil, driverIpmitool)
+	} else {
+		c.Logger.Info("ipmitool provider not available", "error", err.Error())
+	}
 
 	// register ASRR vendorapi provider
-	driverAsrockrack, _ := asrockrack.NewWithOptions(c.Auth.Host, c.Auth.User, c.Auth.Pass, c.Logger, asrockrack.WithHTTPClient(c.httpClient))
+	asrHttpClient := *c.httpClient
```
joelrebel, line 136: I'm speculating here - since `http.Client.Transport` has fields with pointers, and, would it be worth deep copying the client `Transport` with [`Transport.Clone()`](https://pkg.go.dev/net/http#Transport.Clone) ?

joelrebel (commented):
Thanks for all the fixes and changes here, this does make passing parameters to individual providers easier and succinct. I've left a few comments, nothing major.

internal/ipmi/ipmi.go
```diff
@@ -5,42 +5,79 @@ import (
...
+		opt(ipmi)
+	}
+
+	if ipmi.ipmitool == "" {
+		ipmi.ipmitool, err = exec.LookPath("ipmitool")
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		fmt.Println("Using ipmitool path: ", ipmi.ipmitool)
```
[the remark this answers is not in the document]
jacobweinstock: ah thanks, will update

internal/ipmi/ipmi.go
```diff
@@ -5,42 +5,79 @@ import (
...
+		if _, err = os.Stat(ipmi.ipmitool); err != nil {
```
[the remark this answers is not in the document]
jacobweinstock: It's a difference in when the error will occur and it's in line with the existing `exec.LookPath` behavior. With this, the error occurs when `ipmi.New` is called instead of when `Open` or another method is called. This creates a different user experience. 
Walking this thought though, If we remove the `os.Stat` would we also remove the `exec.LookPath`?

option.go
jacobweinstock: this is a good point. i thought about using redfish but opted for gofish. I'm fine with using redfish. Will update.

internal/ipmi/ipmi.go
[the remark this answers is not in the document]
jacobweinstock: done.

option.go
jacobweinstock: done

client.go
jacobweinstock: hmmm....looking into this i say this from the `http.Client` docs. Regardless, i have updated the code to clone the transport.

_"The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines."_

jacobweinstock:
> Thanks for all the fixes and changes here 👍 This does make passing parameters to individual providers easier and succinct. The README requires some updating based on these changes, apart from that I've left a few comments, nothing major.

Thanks, @joelrebel. Could you point me to the needed changes in the README.md, please? I'm not finding anything specific.

internal/ipmi/ipmi.go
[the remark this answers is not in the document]
joelrebel: Oh right I see your point about the user experience, lets leave it as is then.

client.go
joelrebel: ack, feel free to drop the change if its not helping.

joelrebel:
> Thanks, @joelrebel. Could you point me to the needed changes in the README.md, please? I'm not finding anything specific.

sure, in the README we have examples with `bmclib.NewClient()` that currently accept a port as a parameter, which is not required anymore right?

jacobweinstock:
> > Thanks, @joelrebel. Could you point me to the needed changes in the README.md, please? I'm not finding anything specific.
> 
> sure, in the README we have examples with `bmclib.NewClient()` that currently accept a port as a parameter, which is not required anymore right?

Ahh, yes. thank you. I have updated the readme.

client.go
jacobweinstock: i dont know the impacts here either way. i currently have the transport.Clone in there. Which way to you lean?

joelrebel: I'm fine if we go with this for now and see how it works out.

joelrebel (approved):
Thanks for the work here @jacobweinstock

PR #333, opened 2023-05-09
Implements screenshot feature for dell

joelrebel:

Implements screenshot capture feature for Dell IDracs.

### Checklist
- [x] Tests added

Dell

```
- Screenshot feature for Dell Idracs
```

### DIFF
--- a/bmc/screenshot.go
+++ b/bmc/screenshot.go
@@ -1,0 +1,76 @@
+package bmc
+
+import (
+	"context"
+	"fmt"
+
+	bmclibErrs "github.com/bmc-toolbox/bmclib/v2/errors"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+// ScreenshotGetter interface provides methods to query for a BMC screen capture.
+type ScreenshotGetter interface {
+	Screenshot(ctx context.Context) (image []byte, fileType string, err error)
+}
+
+type screenshotGetterProvider struct {
+	name string
+	ScreenshotGetter
+}
+
+// screenshot returns an image capture of the video output.
+func screenshot(ctx context.Context, generic []screenshotGetterProvider) (image []byte, fileType string, metadata Metadata, err error) {
+	var metadataLocal Metadata
+
+	for _, elem := range generic {
+		if elem.ScreenshotGetter == nil {
+			continue
+		}
+		select {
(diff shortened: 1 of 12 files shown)


joelrebel:
rebase on main with changes

bmc/screenshot.go
```diff
@@ -0,0 +1,76 @@
...
+			metadataLocal.SuccessfulProvider = elem.name
+			return image, fileType, metadataLocal, nil
+		}
+	}
+
+	return image, fileType, metadataLocal, multierror.Append(err, errors.New("failed to capture screenshot"))
+}
+
+// ScreenshotFromInterfaces identifies implementations of the ScreenshotGetter interface and passes the found implementations to the screenshot() wrapper method.
+func ScreenshotFromInterfaces(ctx context.Context, generic []interface{}) (image []byte, fileType string, metadata Metadata, err error) {
```
chrisdoherty4, line 52: What needs this signature?

joelrebel: Its how the client passes in the provider registry here - https://github.com/bmc-toolbox/bmclib/blob/49f837d37ccca73f85ca223d214988f36d2417f5/client.go#L364

chrisdoherty4 (approved):
lgtm

jacobweinstock:
Hey @joelrebel , a lot of this looks like duplicate code of the redfish provider. Seems like there are just some specific dell endpoints for being able to screenshot. What would this look like if it was incorporated into the redfish provider? Would it make it too messy?

joelrebel:
@jacobweinstock I did take that approach earlier, and the Redfish provider starts to get [messy](https://github.com/bmc-toolbox/bmclib/blob/main/providers/redfish/tasks.go#L29), which now I'd like to move into the relevant provider. In this case we need to support a minimal set of features that is vendor specific - not present in Redfish spec (and most likely will never be), and so it made more sense have a provider implementation for it.

jacobweinstock:
> @jacobweinstock I did take that approach earlier, and the Redfish provider starts to get [messy](https://github.com/bmc-toolbox/bmclib/blob/main/providers/redfish/tasks.go#L29), which now I'd like to move into the relevant provider. In this case we need to support a minimal set of features that is vendor specific - not present in Redfish spec (and most likely will never be), and so it made more sense have a provider implementation for it.

Ah i see. Thanks!

PR #334, opened 2023-05-25
Screenshot support for Supermicros

joelrebel:

- Implements the Supermicro provider with screen capture support.
- Adds an error type to exclude incompatible provider related errors
- Updates the Dell provider to check for device manufacturer.

### Checklist
- [X] Tests added

Supermicro

SYS-5019C-MR
SYS-510T-MR
6029P-E1CR12L

### DIFF
--- a/client.go
+++ b/client.go
@@ -33,14 +33,15 @@ type Client struct {
 	Logger   logr.Logger
 	Registry *registrar.Registry
 
-	httpClient             *http.Client
-	httpClientSetupFuncs   []func(*http.Client)
-	mdLock                 *sync.Mutex
-	metadata               *bmc.Metadata
-	perProviderTimeout     func(context.Context) time.Duration
-	oneTimeRegistry        *registrar.Registry
-	oneTimeRegistryEnabled bool
-	providerConfig         providerConfig
+	httpClient                       *http.Client
+	httpClientSetupFuncs             []func(*http.Client)
+	mdLock                           *sync.Mutex
+	metadata                         *bmc.Metadata
+	perProviderTimeout               func(context.Context) time.Duration
+	oneTimeRegistry                  *registrar.Registry
+	oneTimeRegistryEnabled           bool
+	returnIncompatibleProviderErrors bool
+	providerConfig                   providerConfig
 }
 
 // Auth details for connecting to a BMC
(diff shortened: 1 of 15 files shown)


client.go
```diff
@@ -32,14 +33,15 @@ type Client struct {
...
-	oneTimeRegistryEnabled bool
-	providerConfig         providerConfig
+	httpClient                       *http.Client
+	httpClientSetupFuncs             []func(*http.Client)
+	mdLock                           *sync.Mutex
+	metadata                         *bmc.Metadata
+	perProviderTimeout               func(context.Context) time.Duration
+	oneTimeRegistry                  *registrar.Registry
+	oneTimeRegistryEnabled           bool
+	returnIncompatibleProviderErrors bool
```
jacobweinstock, line 43: Hey @joelrebel, I like this idea! I had a similar thought a little while back. I created a branch for it but forgot to create the PR. I just opened the PR now. https://github.com/bmc-toolbox/bmclib/pull/335 It's a bit different approach. Let me know what you think. Thanks!

providers/supermicro/supermicro.go
```diff
@@ -0,0 +1,343 @@
+package supermicro
```
jacobweinstock, line 1: There are a few code comments referencing `redfish` in this file. Might want to be `vendorapi`? Maybe copy/paste left over?

jacobweinstock:
I can also confirm that this works on the `SYS-E300-9D-4CN8TP`!

providers/supermicro/supermicro.go
```diff
@@ -0,0 +1,343 @@
...
+)
+
+var (
+	// Features implemented
+	Features = registrar.Features{
+		providers.FeatureScreenshot,
+	}
+)
+
+// supports
```
jacobweinstock, line 43: I can also confirm that this works on the SYS-E300-9D-4CN8TP!

client.go
joelrebel: Hey @jacobweinstock thanks for taking a look, your approach is better in this case and we can include that instead. I'll remove the parts that overlap in my PR.

providers/supermicro/supermicro.go
joelrebel: oh, right, fixed!

client.go
joelrebel: updated in latest commits

providers/supermicro/supermicro.go
```diff
@@ -0,0 +1,343 @@
...
+	host      string
+	user      string
+	pass      string
+	port      string
+	csrfToken string
+	log       logr.Logger
+}
+
+// New returns connection with a Supermicro client initialized
+func NewClient(host, user, pass string, log logr.Logger, opts ...Option) *Client {
```
jacobweinstock, line 88: might think about making `log` optional a then default to `logr.Discard`. Not a PR blocker.

providers/supermicro/supermicro.go
```diff
@@ -0,0 +1,343 @@
...
+	if status != 200 {
+		return errors.Wrap(bmclibErrs.ErrLoginFailed, strconv.Itoa(status))
+	}
+
+	if !bytes.Contains(body, []byte(`url_redirect.cgi?url_name=mainmenu`)) &&
+		!bytes.Contains(body, []byte(`url_redirect.cgi?url_name=topmenu`)) {
+		return errors.Wrap(bmclibErrs.ErrLoginFailed, "unexpected response contents")
+	}
+
+	contentsTopMenu, status, err := c.query(ctx, "cgi/url_redirect.cgi?url_name=topmenu", http.MethodGet, nil, nil, 0)
```
jacobweinstock, line 135: might be helpful to have a code comment about needing to first POST to login and then do a GET for the csrf token. Seems like something i'm going to forget pretty quickly 😬

providers/supermicro/supermicro.go
```diff
@@ -0,0 +1,343 @@
...
+func (c *Client) Screenshot(ctx context.Context) (image []byte, fileType string, err error) {
+	fileType = "jpg"
+
+	// request screen preview to be saved
+	if err := c.initScreenPreview(ctx); err != nil {
+		return nil, "", err
+	}
+
+	// give the bmc a few seconds to store the screen preview
+	time.Sleep(2 * time.Second)
```
jacobweinstock, line 213: ah...this is unfortunate. I imagine this is the best we can do? there's no way programmatically to determine whether the screen preview was stored?

jacobweinstock (approved):
left a few comments, nothing blocking. Thanks @joelrebel. This is a cool feature. I can see it being very useful in a few different ways!

providers/supermicro/supermicro.go
joelrebel: Yeah, theres no means to determine if the is ready. The SMC web interface just delays the fetch for a few seconds.

providers/supermicro/supermicro.go
joelrebel: noted, will include that in the next set of changes

PR #335, opened 2023-05-27
Add metadata field for error messages:

jacobweinstock:
This adds the error message for failed method calls per providers to metadata. These erorrs were previously not accessible. These errors are captured in metadata regardless of whether the method returns successful or not. This is helpful in debugging providers that error but the method call succeeds. This doesn't change the behavior of any methods. After each method call metadata can be interrogated for error messages of any failed providers.
    
The example metadata below shows that while the open method call succeeded for ipmitool, dell, and gofish it did not succeed for IntelAMT and asrockrack. We now have access to the error messages for those failed one.
    
```
{
    SuccessfulProvider:""
    ProvidersAttempted:[ipmitool asrockrack gofish IntelAMT dell]
    SuccessfulOpenConns:[ipmitool dell gofish]
    SuccessfulCloseConns:[]
    FailedConnDetail:
        map[IntelAMT:provider: IntelAMT: unable to perform digest auth with http://192.168.2.205:16992/wsman: Post "http://192.168.2.205:16992/wsman": dial tcp 192.168.2.205:16992: connect: connection refused
            asrockrack:provider: asrockrack: error unmarshalling response payload: invalid character '<' looking for beginning of value
        ]
}
```

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/bmc/power.go
+++ b/bmc/power.go
@@ -40,7 +40,7 @@ type powerProviders struct {
 // setPowerState sets the power state for a BMC, trying all interface implementations passed in
 func setPowerState(ctx context.Context, timeout time.Duration, state string, p []powerProviders) (ok bool, m Metadata, err error) {
 	metadataLocal := Metadata{
-		FailedConnDetail: make(map[string]string),
+		FailedProviderDetail: make(map[string]string),
 	}
 
 	for _, elem := range p {
@@ -59,7 +59,7 @@ func setPowerState(ctx context.Context, timeout time.Duration, state string, p [
 			ok, setErr := elem.powerSetter.PowerSet(ctx, state)
 			if setErr != nil {
 				err = multierror.Append(err, errors.WithMessagef(setErr, "provider: %v", elem.name))
-				metadataLocal.FailedConnDetail[elem.name] = setErr.Error()
+				metadataLocal.FailedProviderDetail[elem.name] = setErr.Error()
 				continue
 			}
 			if !ok {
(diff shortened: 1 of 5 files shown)


bmc/power.go
```diff
@@ -49,14 +51,15 @@ func setPowerState(ctx context.Context, timeout time.Duration, state string, p [
...
-			return false, metadata, err
+			return false, metadataLocal, err
 		default:
 			metadataLocal.ProvidersAttempted = append(metadataLocal.ProvidersAttempted, elem.name)
 			ctx, cancel := context.WithTimeout(ctx, timeout)
 			defer cancel()
 			ok, setErr := elem.powerSetter.PowerSet(ctx, state)
 			if setErr != nil {
 				err = multierror.Append(err, errors.WithMessagef(setErr, "provider: %v", elem.name))
+				metadataLocal.FailedConnDetail[elem.name] = setErr.Error()
```
joelrebel, line 62: Since the error here could be unrelated to the connection Open/Close - maybe we don't include them in `FailedConnDetail` ?

joelrebel (changes requested):
Thanks for this implementation, It would be worth limiting `FailedConnDetail` to just carry Connection Open/Close errors.

bmc/power.go
jacobweinstock: metadata is specific to each method call. So a call to open and the power dont have metadata from each other.

joelrebel: Okay, since this map was called `FailedConnDetail` - I assumed this would only contain errors specific to the connection setup/teardown. But seems like you'd like to include any sort of failure errors in there?

jacobweinstock: Yeah. Maybe I should change the name. Maybe `FailedMethodDetail` or `FailedProviderDetail`?

jacobweinstock: Thanks, @joelrebel. Updated.

joelrebel (approved):

PR #337, opened 2023-07-11
Redfish override timeout for FirmwareInstall

joelrebel:

- Adds a remaining timeout check on the Redfish provider `FirmwareInstall` method.
- Overrides the Redfish wrapper HTTP client timeout for the `FirmwareInstall` method.

The Redfish wrapper client - Gofish has a context timeout on the HTTP client set when `Open()` is invoked, this timeout value is often insufficient for the `FirmwareInstall` method to complete successfully, and so the `FirmwareInstall` method overrides this timeout value and restores it when done.  

### Checklist
- [X] Tests added

```
 - Require a higher context timeout value before proceeding with FirmwareInstall.
 - Set http client timeout to the one passed in through the context to FirmwareInstall.
```

### DIFF
--- a/providers/redfish/firmware.go
+++ b/providers/redfish/firmware.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	gofishrf "github.com/stmcginnis/gofish/redfish"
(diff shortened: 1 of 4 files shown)


providers/redfish/firmware.go
```diff
@@ -43,6 +48,14 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 		return "", errors.Wrap(bmclibErrs.ErrFirmwareInstall, "invalid applyAt parameter: "+applyAt)
 	}
 
+	// expect atleast 10 minutes left in the deadline to proceed with the update
+	//
+	// this gives the BMC enough time to have the firmware uploaded and return a response to the client.
+	ctxDeadline, _ := ctx.Deadline()
+	if time.Until(ctxDeadline) < 10*time.Minute {
+		return "", errors.Wrap(errInsufficientCtxTimeout, " "+time.Until(ctxDeadline).String())
```
DoctorVin, line 56: What's the operational impact of this error? Is this something we expect to have retried? Would we expect a retry to work? Does an exit here leave the BMC in a sane state?

joelrebel: As in with all bmclib functions, its upto the caller to retry with a higher timeout value and/or to close the bmc connection. We provide examples for invoking these methods - https://github.com/bmc-toolbox/bmclib/blob/main/examples/install-firmware/main.go

DoctorVin (approved):

DoctorVin (approved):

PR #339, opened 2023-07-13
supermicro/firmware: support firmware installs on BMCs running older firmware

joelrebel:
support firmware installs on BMCs running firmware <= 1.23.06

The older firmware expects Host headers to be in a certain format (without the port) and it expects a Referrer header. The firmware upload endpoint expects the multipart header slightly different as well.

Tested this change works on the newer firmware (1.74.x) as well.

### Checklist
- [X] Tests added

- SMC X11s

```
- Support for SMC firmware updates on BMCs running firmware <=1.23.06
```


diogomatsubara (approved):
LGTM. Thanks for the fix!

PR #340, opened 2023-07-19
SEK-322 Sol and ciphers

nuclearbob:
IPMI functions for sol and getciphers
Example files demonstrating usage of those functions
### Checklist
I intend to write tests that use these functions, but I have not written tests for the functions themselves. If there is a good place to add those, I can.

```
IPMI sol and getciphers functions added for use in hardware testing
```

### DIFF
--- a/examples/getCiphers/getCiphers.go
+++ b/examples/getCiphers/getCiphers.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/bmc-toolbox/bmclib/v2/providers/ipmitool"
+	"github.com/bombsimon/logrusr/v2"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	// set BMC parameters here
+	host := "10.211.132.157"
+	user := "root"
+	pass := "yxvZdxAQ38ZWlZ"
+
+	l := logrus.New()
+	l.Level = logrus.DebugLevel
+	logger := logrusr.New(l)
+
+	if host == "" || user == "" || pass == "" {
+		log.Fatal("required host/user/pass parameters not defined")
+	}
+
+	i, err := ipmitool.New(host, user, pass, ipmitool.WithLogger(logger))
+	if err != nil {
+		log.Fatal("ipmi connection failed")
+	}
+
+	err = i.Open(ctx)
+	if err != nil {
+		log.Fatal(err, "ipmi login failed")
+	}
+
+	defer i.Close(ctx)
+
+	output, err := i.GetIPMICiphers(ctx)
+	if err != nil {
+		l.Error(err)
+	}
+	log.Print("GetIPMICiphers returned ", output)
+
+	output, err = i.GetSOLCiphers(ctx)
+	if err != nil {
+		l.Error(err)
+	}
+	log.Print("GetSOLCiphers returned ", output)
+
+}
--- a/examples/sol/sol.go
+++ b/examples/sol/sol.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/bmc-toolbox/bmclib/v2/providers/ipmitool"
+	"github.com/bombsimon/logrusr/v2"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	// set BMC parameters here
+	host := "10.211.132.157"
+	user := "root"
+	pass := "yxvZdxAQ38ZWlZ"
+
+	l := logrus.New()
+	l.Level = logrus.DebugLevel
+	logger := logrusr.New(l)
+
+	if host == "" || user == "" || pass == "" {
+		log.Fatal("required host/user/pass parameters not defined")
+	}
+
+	i, err := ipmitool.New(host, user, pass, ipmitool.WithLogger(logger))
+	if err != nil {
+		log.Fatal("ipmi connection failed")
+	}
+
+	err = i.Open(ctx)
+	if err != nil {
+		log.Fatal(err, "ipmi login failed")
+	}
+
+	defer i.Close(ctx)
+
+	info, err := i.SolInfo(ctx)
+	if err != nil {
+		l.Error(err)
+	}
+	log.Print("SolInfo returned ", info)
+
+	cancel()
+	ctx, cancel = context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+	go func() {
+		time.Sleep(5 * time.Second)
+		i.SolDeactivate(ctx)
+	}()
+
+	// We expect an error here
+	output, _ := i.SolActivate(ctx, []byte("\n\n")...)
+	log.Print("SolActivate returned ", output)
+}
--- a/internal/ipmi/ipmi.go
+++ b/internal/ipmi/ipmi.go
@@ -71,7 +71,7 @@ func New(username string, password string, host string, opts ...Option) (ipmi *I
 	return ipmi, err
 }
 
-func (i *Ipmi) run(ctx context.Context, command []string) (output string, err error) {
+func (i *Ipmi) run(ctx context.Context, command []string, stdin ...byte) (output string, err error) {
 	var out []byte
 	var ipmiCiphers = []string{"3", "17"}
 	ipmiArgs := []string{"-I", "lanplus", "-U", i.Username, "-E", "-N", "5"}
@@ -93,6 +93,13 @@ func (i *Ipmi) run(ctx context.Context, command []string) (output string, err er
 		i.log.V(3).Info("ipmitool options", "opts", formatOptions(ipmiCmd))
 		ipmiCmd = append(ipmiCmd, command...)
 		cmd := exec.CommandContext(ctx, i.ipmitool, ipmiCmd...)
+		if stdin != nil {
+			stdinPipe, err := cmd.StdinPipe()
+			if err != nil {
+				return "", err
+			}
+			stdinPipe.Write(stdin)
+		}
 		cmd.Env = []string{fmt.Sprintf("IPMITOOL_PASSWORD=%s", i.Password)}
 		out, err = cmd.CombinedOutput()
 		if err == nil || ctx.Err() != nil {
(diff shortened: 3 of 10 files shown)


examples/getCiphers/getCiphers.go
```diff
@@ -0,0 +1,54 @@
+package main
```
joelrebel, line 1: since this example calls into the ipmitool provider directly, I suggest naming the parent directory, `ipmitool-getCiphers`

examples/getCiphers/getCiphers.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/bmc-toolbox/bmclib/v2/providers/ipmitool"
+	"github.com/bombsimon/logrusr/v2"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	// set BMC parameters here
```
joelrebel, line 17: I suggest using the same format as the other [examples]( https://github.com/bmc-toolbox/bmclib/tree/main/examples/inventory)

- Add a doc.go with some notes about the example.
- Add an output.txt or such with the output returned by this example.
- Accept flags instead for parameters.

examples/sol/sol.go
```diff
@@ -0,0 +1,59 @@
+package main
```
joelrebel, line 1: follow the suggestions in the previous comments for this as well

internal/ipmi/ipmi.go
```diff
@@ -377,3 +384,70 @@ func (i *Ipmi) ReadUsers(ctx context.Context) (users []map[string]string, err er
...
 }
+
+// SolInfo gets the status of the serial-over-lan interface
+func (i *Ipmi) SolInfo(ctx context.Context) (info string, err error) {
+	output, err := i.run(ctx, []string{"sol", "info"})
+	if err != nil {
+		return "", fmt.Errorf("%v: %v", err, output)
+	}
+
+	fmt.Println(output)
```
joelrebel, line 395: remove debug prints

internal/ipmi/ipmi.go
```diff
@@ -377,3 +384,70 @@ func (i *Ipmi) ReadUsers(ctx context.Context) (users []map[string]string, err er
...
+
+// SolActivate activates and connects to the serial-over-lan interface.
+// It needs stdin, or the connection will be immediately dropped.
+func (i *Ipmi) SolActivate(ctx context.Context, stdin ...byte) (output string, err error) {
+	output, err = i.run(ctx, []string{"sol", "activate"}, stdin...)
+	if err != nil {
+		return output, fmt.Errorf("%v: %v", err, output)
+	}
+
+	if strings.HasPrefix(output, "[SOL Session operational.") {
```
joelrebel, line 409: Is this string consistent across vendors, models (atleast the ones we care about)?

internal/ipmi/ipmi.go
```diff
@@ -377,3 +384,70 @@ func (i *Ipmi) ReadUsers(ctx context.Context) (users []map[string]string, err er
...
+
+// SolDeactivate disconnects the serial-over-lan interface.
+func (i *Ipmi) SolDeactivate(ctx context.Context) (output string, err error) {
+	output, err = i.run(ctx, []string{"sol", "deactivate"})
+	if err != nil {
+		return output, fmt.Errorf("%v: %v", err, output)
+	}
+
+	// If we see this, it was already deactivated, return as error.
+	if strings.HasPrefix(output, "Info: SOL payload already de-activated") {
```
joelrebel, line 423: I don't think this should return an error, since from the callers perspective no further action can/needs to be taken.

internal/ipmi/ipmi.go
```diff
@@ -377,3 +384,70 @@ func (i *Ipmi) ReadUsers(ctx context.Context) (users []map[string]string, err er
...
+		return output, fmt.Errorf("%v: %v", err, output)
+	}
+	return output, nil
+}
+
+// GetIPMICiphers gets a list of ciphers supported for IPMI.
+func (i *Ipmi) GetIPMICiphers(ctx context.Context) (output string, err error) {
```
joelrebel, line 430: I suggest calling this just `GetCiphers` since it'll be invoked as `ipmi.GetCiphers` and the name does not stutter.

internal/ipmi/ipmi.go
```diff
@@ -377,3 +384,70 @@ func (i *Ipmi) ReadUsers(ctx context.Context) (users []map[string]string, err er
...
+	output, err = i.run(ctx, []string{"channel", "getciphers", "ipmi"})
+	if err != nil {
+		return output, fmt.Errorf("%v: %v", err, output)
+	}
+
+	if strings.HasPrefix(output, "ID   IANA") {
```
joelrebel, line 436: Is this string consistent across vendors?

joelrebel (changes requested):
This seems fine to me, I've left a few comments inline. 
It would be worth testing on multiple hardware to verify certain string comparisons work consistently.

internal/ipmi/ipmi.go
nuclearbob: For testing purposes I'd like to know if the deactivate I called succeeded or if something else deactivated the session first. I can do that by returning the output instead of returning an error. 👍

internal/ipmi/ipmi.go
nuclearbob: I'll check on this one and the other one across a wider set of hardware today using the test tool you linked.

examples/getCiphers/getCiphers.go
nuclearbob: I modeled mine after reset_bmc.go. I'll rework it to model after inventory instead, thanks!

nuclearbob:
> This seems fine to me, I've left a few comments inline. It would be worth testing on multiple hardware to verify certain string comparisons work consistently.

I'm working on the changes you requested. I noticed several checks failed as well. I think I addressed the lint issue, but I'm not sure what to do about codecov. Is there a way to write the examples as tests, or do tests need to be written separately, or is that something I don't need to pay attention to yet?

internal/ipmi/ipmi.go
nuclearbob: I tested this string and the cipher strings across g3 plans with hardware from dell, asrockrack, smc, and gigabyte. All of them matched. The tests using these features will be deployed more widely against the fleet, so if we find servers with different behavior, we can collect that info, but so far they all seem to match.

providers/ipmitool/ipmitool.go
```diff
@@ -180,3 +180,29 @@ func (c *Conn) PowerSet(ctx context.Context, state string) (ok bool, err error)
...
+
+// GetCiphers gets a list of ciphers supported for IPMI.
+func (c *Conn) GetCiphers(ctx context.Context) (output string, err error) {
+	return c.ipmitool.GetCiphers(ctx)
+}
+
+// GetSOLCiphers gets a list of ciphers supported for SOL.
+func (c *Conn) GetSOLCiphers(ctx context.Context) (output string, err error) {
+	return c.ipmitool.GetSOLCiphers(ctx)
+}
```
jacobweinstock, line 208: Hey @joelrebel, im concerned that this doesnt follow the model of abstracting providers. I think if we want to support these they should become interface definition in the `bmc` package. Especially because, i believe, `racadm` can support SOL and providers already have to opt in to functionality.

Also, I'm not convinced that `SolInfo`, `GetCiphers`, `GetSOLCiphers` warrant first class interface methods. Maybe @nuclearbob you can share some more detail about that these methods would allow you to accomplish? For example, if you were wanted to be able to get the Ciphers available and then use one of them for another call. In this scenario we might benefit more from integrating that discovering internally to the ipmitool implementation instead of exposing it via our public API. 

Regardless, thanks for the PR!

nuclearbob: I'm writing tests to validate the behavior of different bmc hardware. Some of the things we want to check are these specific IPMI functions. bmclib is our preferred method of interfacing with a bmc in go, so this was a simple way to add that functionality. For my purposes, an externally available method to run an arbitrary ipmitool command would work fine, so if that's workable, I can propose it instead. If you have a different preference, like implementing SOL functionality that can also be used in the racadm provider, let me know. Whatever makes the most sense for the library works for me.

jacobweinstock: Hey @nuclearbob , thanks for sharing your use-case. Unfortunately, "an externally available method to run an arbitrary ipmitool command" is not something i support for bmclib. This would create a tight coupling with ipmitool and significantly degrade our abstraction layer. For generic ipmitool functionality in Go, i would recommend using the stdlib `"os/exec"` to call directly to `ipmitool`.  So with all that said, i would recommend the SOL functionality be added via an interface in the `bmc` package.

Would love to get @joelrebel 's thoughts here too.

For what its worth, this idea was brought up in the Tinkerbell PBnJ repo as well: https://github.com/tinkerbell/pbnj/issues/26

joelrebel: @nuclearbob Jacob makes a fair point here, let's have this functionality exposed via the interface in the `bmc` package. With this done, bmclib-tester can be used to test them as a result.

To expose the methods exposed via the `bmc` interface, follow the steps below
the links point to existing use cases.

1. Add the interface wrapper methods - https://github.com/bmc-toolbox/bmclib/blob/main/bmc/postcode.go
2. Enlist them in the bmc client, for example - https://github.com/bmc-toolbox/bmclib/blob/a1b87e2ff47f5c05022a6cb8fddc6811e13d9a89/client.go#L369
3. Define each of the feature you're adding as a constant here https://github.com/bmc-toolbox/bmclib/blob/a1b87e2ff47f5c05022a6cb8fddc6811e13d9a89/providers/providers.go#L33 (`SerialOverLanActivate`...)
4. Define the added features in the ipmitool provider https://github.com/bmc-toolbox/bmclib/blob/a1b87e2ff47f5c05022a6cb8fddc6811e13d9a89/providers/ipmitool/ipmitool.go#L22

Suggestion: can the methods be prefixed by `SoL` instead of `Sol` or `SerialOverLan` for clarity.

nuclearbob: Will #1 require tests to be written, like https://github.com/bmc-toolbox/bmclib/blob/main/bmc/postcode_test.go for your example? If so, we may be exceeding the scope of the motivating issue on my team. I'll get started on the PR and discuss with my supervisor.

joelrebel: @nuclearbob `SerialOverLan` would be preferred,
 The tests are mostly boilerplate code (we need to look into having them generated at some point), either way, I can help implement the tests if you can add the rest of the changes.

nuclearbob:
The scope on this got a little big for what we really wanted to do (basic output verification of 5 ipmitool calls) so we implemented an alternate connector for the testsuite. We may revisit the feature addition later.

PR #342, opened 2023-08-09
Support unstructured HTTP update

ofaurax:
Even if deprecated, this PR adds support for the "unstructured HTTP update" from Redfish specification.
The current implementation only support multipart update.
This PR adds a fallback to "unstructured HTTP update" as a fallback of HTTP multipart is not available.

### Checklist
- [x] Tests added

OpenBMC redfish implementation

```
Initial release
```


providers/redfish/firmware.go
```diff
@@ -106,7 +106,23 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 		"UpdateFile":       reader,
 	}
 
-	resp, err := c.runRequestWithMultipartPayload(http.MethodPost, "/redfish/v1/UpdateService/MultipartUpload", payload)
+	updateService, err := c.redfishwrapper.UpdateService()
```
joelrebel, line 109: I suggest we update the `firmwareUpdateCompatible()` with the logic below, to return the update URI and it returns an error if the URI is not supported.

providers/redfish/firmware.go
```diff
@@ -124,6 +140,13 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 		taskID = strings.Split(resp.Header.Get("Location"), "JID_")[1]
 	}
```
joelrebel, line 142: lets move both these lookups for the taskID into its own function

providers/redfish/firmware.go
```diff
@@ -138,6 +161,47 @@ func (c *Conn) FirmwareInstallStatus(ctx context.Context, installVersion, compon
 	switch {
 	case strings.Contains(vendor, constants.Dell):
 		task, err = c.dellJobAsRedfishTask(taskID)
+	case strings.Contains(vendor, constants.Packet):
+		resp, _ := c.redfishwrapper.Get("/redfish/v1/TaskService/Tasks/" + taskID)
```
joelrebel, line 165: You might be able to use the TaskService method from Gofish  - https://github.com/stmcginnis/gofish/blob/6e8a2909577b8424e751890a5c432352f227097b/serviceroot.go#L257

providers/redfish/firmware.go
```diff
@@ -149,11 +213,13 @@ func (c *Conn) FirmwareInstallStatus(ctx context.Context, installVersion, compon
 		return state, err
 	}
 
-	if task == nil {
-		return state, errors.New("failed to lookup task status for task ID: " + taskID)
-	}
+	if state == "" {
```
joelrebel, line 216: why is this check required?

providers/redfish/firmware.go
```diff
@@ -248,6 +315,16 @@ func (c *Conn) runRequestWithMultipartPayload(method, url string, payload map[st
 	return c.redfishwrapper.RunRawRequestWithHeaders(method, url, bytes.NewReader(payloadBuffer.Bytes()), payloadWriter.FormDataContentType(), nil)
 }
 
+func (c *Conn) runRequestWithPayload(method, url string, payload io.Reader) (*http.Response, error) {
+	if url == "" {
+		return nil, fmt.Errorf("unable to execute request, no target provided")
+	}
+
+	b, _ := io.ReadAll(payload)
```
joelrebel, line 323: I'd like to see if we have other options than reading the whole file into memory

joelrebel (commented):
Thanks for the efforts here, I've left a few suggestions inline

providers/redfish/firmware.go
ofaurax: yes, done.

providers/redfish/firmware.go
ofaurax: Previous work, irrelevant now

providers/redfish/firmware.go
ofaurax: The `payloadReadSeeker` needs to be of type `io.ReadSeeker` meaning we need to be able to read and seek anywhere in the data, so we're forced to have it all.

On the multipart update, I see `bytes.NewReader(payloadBuffer.Bytes())`.
I'm not sure if the whole update is not in memory at that time.

constants/constants.go
```diff
@@ -22,8 +22,11 @@ const (
 	Cloudline = "Cloudline"
 	// Quanta is the contant to identify Quanta hardware
 	Quanta = "Quanta"
-	// Quanta is the contant to identify Intel hardware
+	// Intel is the contant to identify Intel hardware
 	Intel = "Intel"
+	// Equinix, formerly Packet
```
jacobweinstock, line 27: what are these identifying? does Packet and Equinix have hardware like Intel and or Quanta do?

jacobweinstock: maybe this is related to a customer openBMC build Equinix/Packet are running with redfish in it? If this is the case we should make that clear here. Without it I'm left to assume this is a server that built by Equinix/Packet like intel or dell.

providers/redfish/firmware.go
joelrebel: Thanks for pointing this out, I realised that this was a regression and its supposed to be a stream upload https://github.com/bmc-toolbox/bmclib/pull/345

constants/constants.go
joelrebel: In this case the hardware was manufactured with the vendor attribute set to `Packet`.

joelrebel:
The work on this PR was moved into https://github.com/bmc-toolbox/bmclib/pull/346 since we reworked the implementation.

PR #343, opened 2023-08-12
RPC provider

jacobweinstock:

This adds an RPC provider. The idea of the RPC provider is to enable users the ability to interoperate with existing bespoke out of band system or as a stop gap for BMCs/protocols that bmclib does not have a provider for yet. Internal bespoke out of band systems will generally not be accepted into BMCLIB.

So users can run a small web endpoint that implements the RPC contract. The RPC provider sends a request to the user's web endpoint. The web endpoint will then translate the call to the out of band protocol/tool/etc that is needed to communicate with the BMC.

### Checklist
- [ ] Tests added

```
```


providers/rpc/doc.go
```diff
@@ -0,0 +1,8 @@
+/*
+Package rpc is a provider that defines an HTTP request/response contract for handling BMC interactions.
+It allows users a simple way to interoperate with an existing/bespoke out-of-band management solution.
+
+The rpc provider request/response payloads are modeled after JSON-RPC 2.0, but are not JSON-RPC 2.0
```
joelrebel, line 5: Out of curiousity, 
 - Why pick JSON-RPC in the first place, instead of lets say https://github.com/twitchtv/twirp
 - If JSON-RPC seems ideal for this usecase, could we not use an existing library https://github.com/sourcegraph/jsonrpc2
 
I'm a bit wary of having to maintain our own JSON-RPC implementation.

jacobweinstock: ah yeah, very fair question. Flexibility is the biggest factor. I have found being conformant to a spec as a barrier to interoperating with bespoke out-of-band systems. I shied away from any protocol buffers because of the user experience for creating clients and interacting with the provider. I did look into using https://github.com/sourcegraph/jsonrpc2. The biggest reason i did not go with it was I couldn't find a way to add headers to the calls. Having some sort of authn/authz was important, in my experience, and i just couldn't find a way to add auth to headers in this library. And secondary, it meant that we would be using the strict JSON-RPC spec.

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+	SHA256 Algorithm = "sha256"
+	// SHA256Short is the short version of the SHA256 algorithm.
+	SHA256Short Algorithm = "256"
+	// SHA512 is the SHA512 algorithm.
+	SHA512 Algorithm = "sha512"
+	// SHA512Short is the short version of the SHA512 algorithm.
+	SHA512Short Algorithm = "512"
+)
+
+// Features implemented by the AMT provider.
```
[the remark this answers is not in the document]
jacobweinstock: ah thanks, fix pushed.

providers/rpc/doc.go
jacobweinstock: I agree i dont want to maintain our own JSON-RPC implementation. I think at the moment the contract we have is just a request/response and it's about as simple/basic as i could think. It's HTTP and its general constructs like body and headers.

joelrebel: Thanks for the explanation Jacob, as long we keep the foot print small it should be maintainable. Are you able to include a consumer example for reference in the examples/ directory?

providers/rpc/experimental.go
```diff
@@ -0,0 +1,27 @@
+package rpc
+
+import (
+	"github.com/Jeffail/gabs/v2"
+	"github.com/ghodss/yaml"
+)
+
+// embedPayload will embed the RequestPayload into the given JSON object at the dot path notation location ("object.data").
+func (p *RequestPayload) embedPayload(rawJSON []byte, dotPath string) ([]byte, error) {
+	if rawJSON != nil {
```
joelrebel, line 10: this could return early by inverting the check 
```
if rawJSON == nil {
  return ...
}
```

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+		return err
+	}
+	c.listenerURL = u
+	testReq, err := http.NewRequestWithContext(ctx, c.Opts.Request.HTTPMethod, c.listenerURL.String(), nil)
+	if err != nil {
+		return err
+	}
+	// test that we can communicate with the rpc consumer.
+	// and that it responses with the spec contract (Response{}).
+	if _, err := c.Opts.Request.Client.Do(testReq); err != nil { //nolint:bodyclose // not reading the body
```
joelrebel, line 188: would be ideal the whole body and close - so the connection can be reused (no?)

https://pkg.go.dev/net/http#Client.Do
```
If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.
```

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+
+	resp, err := c.Opts.Request.Client.Do(req)
+	if err != nil {
+		c.Logger.Error(err, "failed to send rpc notification", kvs...)
+		return ResponsePayload{}, err
+	}
+	defer resp.Body.Close()
+
+	// handle the response
+	rp, err := c.handleResponse(resp, kvs)
```
joelrebel, line 327: Im guessing at this point the consumer is trusted, how is this trust established?

providers/rpc/http.go
```diff
@@ -0,0 +1,77 @@
...
+func (c *Config) handleResponse(resp *http.Response, reqKeysAndValues []interface{}) (ResponsePayload, error) {
+	kvs := reqKeysAndValues
+	defer func() {
+		if !c.LogNotificationsDisabled {
+			kvs = append(kvs, responseKVS(resp)...)
+			c.Logger.Info("rpc notification details", kvs...)
+		}
+	}()
+	defer resp.Body.Close()
+	bodyBytes, err := io.ReadAll(resp.Body)
```
joelrebel, line 57: Should there be limit/check for the response body size, just to make sure the provider isn't vulnerable to being dos'ed ?

client.go
```diff
@@ -130,6 +134,21 @@ func (c *Client) defaultTimeout(ctx context.Context) time.Duration {
 }
 
 func (c *Client) registerProviders() {
+	// register the rpc provider
```
joelrebel, line 137: since the `registerProviders()` method is getting longer, for readability, it would ideal to move these into smaller functions - `(c *Client) registerRPCProvider() error`

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+type SignatureOpts struct {
+	// HeaderName is the header name that should contain the signature(s). Example: X-BMCLIB-Signature
+	HeaderName string
+	// AppendAlgoToHeaderDisabled decides whether to append the algorithm to the signature header or not.
+	// Example: X-BMCLIB-Signature becomes X-BMCLIB-Signature-256
+	// When set to true, a header will be added for each algorithm. Example: X-BMCLIB-Signature-256 and X-BMCLIB-Signature-512
+	AppendAlgoToHeaderDisabled bool
+	// IncludedPayloadHeaders are headers whose values will be included in the signature payload. Example: X-BMCLIB-My-Custom-Header
+	// All headers will be deduplicated.
+	IncludedPayloadHeaders []string
```
chrisdoherty4, line 112: How would this get used? Based on the example, it seems like a way to add arbitrary hedaers but I see we also have `StaticHeaders` elsewhere.

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+	TimestampHeader string
+}
+
+type SignatureOpts struct {
+	// HeaderName is the header name that should contain the signature(s). Example: X-BMCLIB-Signature
+	HeaderName string
+	// AppendAlgoToHeaderDisabled decides whether to append the algorithm to the signature header or not.
+	// Example: X-BMCLIB-Signature becomes X-BMCLIB-Signature-256
+	// When set to true, a header will be added for each algorithm. Example: X-BMCLIB-Signature-256 and X-BMCLIB-Signature-512
+	AppendAlgoToHeaderDisabled bool
```
chrisdoherty4, line 109: What's the motivation for allowing this to be disabled?

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+	Signature SignatureOpts
+	// HMAC is the options used to create a HMAC signature.
+	HMAC HMACOpts
+	// Experimental options.
+	Experimental Experimental
+}
+
+type RequestOpts struct {
+	// Client is the http client used for all HTTP calls.
+	Client *http.Client
```
chrisdoherty4, line 90: It feels a bit peculiar to have a client nested here as opposed to on the top `Config` object. I appreciate its related to HTTP but as far as I can tell its not really used to _construct_ a request, its used to _submit_ it.

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+	// IncludedPayloadHeaders are headers whose values will be included in the signature payload. Example: X-BMCLIB-My-Custom-Header
+	// All headers will be deduplicated.
+	IncludedPayloadHeaders []string
+}
+
+type HMACOpts struct {
+	// Hashes is a map of algorithms to a slice of hash.Hash (these are the hashed secrets). The slice is used to support multiple secrets.
+	Hashes map[Algorithm][]hash.Hash
+	// PrefixSigDisabled determines whether the algorithm will be prefixed to the signature. Example: sha256=abc123
+	PrefixSigDisabled bool
```
chrisdoherty4, line 119: Similar to other comments, whats the motivation to provide this level of flexibility?

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+
+// Secrets hold per algorithm slice secrets.
+// These secrets will be used to create HMAC signatures.
+type Secrets map[Algorithm][]string
+
+// Signatures hold per algorithm slice of signatures.
+type Signatures map[Algorithm][]string
+
+// Config defines the configuration for sending rpc notifications.
+type Config struct {
```
chrisdoherty4, line 60: It seems peculiar that a data structure named `Config` would provide the various behaviors in this PR. At the consumer site that means we `rpc.Config` and pass that around. `rpc.Provider` seems more representative for example. 

I don't wish to disrupt convention but thought I'd call it out nonetheless.

providers/rpc/logging.go
```diff
@@ -0,0 +1,61 @@
...
+// requestKVS returns a slice of key, value sets. Used for logging.
+func requestKVS(req *http.Request) []interface{} {
+	var r requestDetails
+	if req != nil && req.Body != nil {
+		var p RequestPayload
+		reqBody, err := io.ReadAll(req.Body)
+		if err == nil {
+			req.Body = io.NopCloser(bytes.NewBuffer(reqBody))
+			_ = json.Unmarshal(reqBody, &p)
+		}
```
chrisdoherty4, line 32: I believe there's a leak here because the underlying body is never closed on success, its just switched out with the `io.NopCloser()`.

providers/rpc/payload.go
```diff
@@ -0,0 +1,69 @@
...
+	PowerGetMethod     Method = "getPowerState"
+	VirtualMediaMethod Method = "setVirtualMedia"
+)
+
+// RequestPayload is the payload sent to the ConsumerURL.
+type RequestPayload struct {
+	ID     int64       `json:"id"`
+	Host   string      `json:"host"`
+	Method Method      `json:"method"`
+	Params interface{} `json:"params,omitempty"`
```
chrisdoherty4, line 19:
```suggestion
	Params any `json:"params,omitempty"`
```

?

providers/rpc/payload.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// ResponsePayload is the payload received from the ConsumerURL.
+// The Result field is an interface{} so that different methods
+// can define the contract according to their needs.
+type ResponsePayload struct {
+	// ID is the ID of the response. It should match the ID of the request but is not enforced.
+	ID     int64          `json:"id"`
+	Host   string         `json:"host"`
+	Result interface{}    `json:"result,omitempty"`
```
chrisdoherty4, line 47:
```suggestion
	Result any    `json:"result,omitempty"`
```

providers/rpc/http.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (c *Config) handleResponse(resp *http.Response, reqKeysAndValues []interface{}) (ResponsePayload, error) {
+	kvs := reqKeysAndValues
+	defer func() {
+		if !c.LogNotificationsDisabled {
+			kvs = append(kvs, responseKVS(resp)...)
+			c.Logger.Info("rpc notification details", kvs...)
+		}
+	}()
+	defer resp.Body.Close()
```
chrisdoherty4, line 56: I don't recall how defer works but are you sure this closes the right handle given we reset the body on l.74?

(it may be the case that Go stores the point in memory it'll jump to on return at the time of `defer`, hence I'm not sure if this is a problem)

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+		return false, fmt.Errorf("error from rpc consumer: %v", rp.Error)
+	}
+
+	return true, nil
+}
+
+// PowerSet sets the power state of a BMC machine.
+func (c *Config) PowerSet(ctx context.Context, state string) (ok bool, err error) {
+	switch strings.ToLower(state) {
+	case "on", "off", "cycle":
```
chrisdoherty4, line 226: I assume we don't have a concrete type for the states (enum style)?

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+		return ResponsePayload{}, err
+	}
+
+	// create the signature payload
+	// get the body and reset it as readers can only be read once.
+	body, err := io.ReadAll(req.Body)
+	if err != nil {
+		return ResponsePayload{}, err
+	}
+	req.Body = io.NopCloser(bytes.NewBuffer(body))
```
chrisdoherty4, line 286: Similar to my comment in the `requestKVS` function, I think this is a leak.

chrisdoherty4 (commented):
Nice to see this shaping up.

I noticed there's a fair amount of flexibility and I'm not sure I appreciate why. For example, there are options to specify the HTTP content type and HTTP method used: whats the motivation for being so flexible when we are ultimately defining the contract?

In the case of content type, given we assume JSON does it make sense to even expose that in such a raw manner? (this makes me wonder if I really understand whats going on 🤔).

providers/rpc/rpc.go
chrisdoherty4: I think this is a sort of filter that specifies which headers are included in the payload that gets signed?

The naming through me, but I think that's a me thing. Perhaps the example could be elaborated on 🤷🏻

providers/rpc/http.go
jacobweinstock: sure. Any suggestions on how to define an upper limit? The `ResponsePayload.Result` is of type `interface{}` so to allow for different methods to return the types they need. This field could be of variable length. And then we have some string fields that could be large or small depending on the consumer.

jacobweinstock:
> Nice to see this shaping up.
> 
> I noticed there's a fair amount of flexibility and I'm not sure I appreciate why. For example, there are options to specify the HTTP content type and HTTP method used: whats the motivation for being so flexible when we are ultimately defining the contract?
> 
> In the case of content type, given we assume JSON does it make sense to even expose that in such a raw manner? (this makes me wonder if I really understand whats going on 🤔).

Thanks for the review, Chris. 

_whats the motivation for being so flexible when we are ultimately defining the contract?_

With the goal of the RPC provider being to provide interoperability with existing/bespoke out-of-band systems, for me, flexibility is key. `POST` vs `PATCH` or even a `GET` for the HTTP method will provide that flexibility while not compromising our payload contracts. The content type, while we will always send json, could have a requirement in the existing system. i.e. `application/vnd.api+json` or `application/json` or some other custom type.

providers/rpc/rpc.go
jacobweinstock: This is in the `SignatureOpts` and pertains only the the creation of the HMAC hash. These header names/keys must exist before hand. This option will not add headers to the request.

providers/rpc/rpc.go
jacobweinstock: sure, no problem. i will move it to the top level in `Config`.

providers/rpc/rpc.go
jacobweinstock: yeah, fair. the motivation is interoperability with systems where HMAC might have a specific requirement with regard to its payload signature header. I want to provide as much flexibility for users and bmclib will most likely be the new piece of the puzzle that will most likely have to be more flexible in order to interoperate properly.

providers/rpc/rpc.go
jacobweinstock: no problem. i will change it. it was honestly not something i put much thought into so im glad to get your feedback.

providers/rpc/rpc.go
jacobweinstock: correct. we do not.

providers/rpc/rpc.go
jacobweinstock: in terms of the example, what do you think would improve the clarity of this field? maybe something like,

 "_given these headers in a request, the keys specified in `IncludedPayloadHeaders` will take the values of those keys and include them in the signature payload_"?

providers/rpc/payload.go
jacobweinstock: updated.

providers/rpc/payload.go
jacobweinstock: updated

providers/rpc/rpc.go
jacobweinstock: sorry, dont know if i follow. mind elaborating?

providers/rpc/logging.go
chrisdoherty4: To elaborate. This code reads the body on l.28, then if no error occurs proceeds to swap the body with a `NopCloser()` without closing the `req.Body`'s original `io.ReadCloser`. Consequently, we risk a leak, or in the least inhibit the underlying TCP connection from being reused as per the `http.Request` docs.

I mention the leak because I've observed not calling `req.Body.Close()` result in OS file descriptor (there's even articles dedicated to the problem https://www.j4mcs.dev/posts/golang-response-body/).

chrisdoherty4: 🤦🏻 I've just realized by brain fart, this is a Request, not a Reponse ...

providers/rpc/rpc.go
jacobweinstock: updated

providers/rpc/http.go
```diff
@@ -0,0 +1,77 @@
...
+		req.Header.Set("Content-Type", c.Opts.Request.HTTPContentType)
+	}
+	if c.Opts.Request.TimestampHeader != "" {
+		req.Header.Add(c.Opts.Request.TimestampHeader, time.Now().Format(c.Opts.Request.TimestampFormat))
+	}
+
+	return req, nil
+}
+
+func (c *Config) handleResponse(resp *http.Response, reqKeysAndValues []interface{}) (ResponsePayload, error) {
```
chrisdoherty4:
```suggestion
func (c *Config) handleResponse(resp *http.Response, reqKeysAndValues []any) (ResponsePayload, error) {
```

providers/rpc/rpc.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	defer resp.Body.Close()
+
+	// handle the response
+	rp, err := c.handleResponse(resp, kvs)
+	if err != nil {
+		return ResponsePayload{}, err
+	}
+
+	return rp, nil
```
chrisdoherty4:
```suggestion
	// handle the response
	return c.handleResponse(resp, kvs)
```

providers/rpc/http.go
jacobweinstock: Hey @joelrebel , i've added `maxContentLenAllowed` and a content length check. The response contract we've defined is pretty small and my tests showed that currently only about 50 bytes are needed. With that in mind i made `maxContentLenAllowed` equal to 512KB. Thoughts?

providers/rpc/rpc.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	resp, err := p.Client.Do(req)
+	if err != nil {
+		p.Logger.Error(err, "failed to send rpc notification", kvs...)
+		return ResponsePayload{}, err
+	}
+	defer resp.Body.Close()
+	respBuf := new(bytes.Buffer)
+	if _, err := io.Copy(respBuf, resp.Body); err != nil {
```
chrisdoherty4, line 337: I think this may still be at risk of a DoS because `Copy` will read everything from `Body` into the buffer. I think, in theory, that could consume a CPU but also chew threw all available memory.

I think the only way to safely read is to chunk (https://pkg.go.dev/io#CopyN).

providers/rpc/rpc.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	// request/response round trip.
+	kvs := requestKVS(req)
+	kvs = append(kvs, []interface{}{"host", p.Host, "method", rp.Method, "consumerURL", p.ConsumerURL}...)
+	if rp.Params != nil {
+		kvs = append(kvs, []interface{}{"params", rp.Params}...)
+	}
+
+	resp, err := p.Client.Do(req)
```
chrisdoherty4, line 330: Do we specify a timeout on the request?

providers/rpc/rpc.go
jacobweinstock: ah yeah, fair. i'll use `CopyN` and move the copy below the content length check.

providers/rpc/rpc.go
jacobweinstock: We do. We use `http.NewRequestWithContext` to create the `req`.

joelrebel (approved):
Thanks for the efforts here 🚀

providers/rpc/rpc.go
chrisdoherty4: Yea OK. I was thinking about affording them some protection from malicious/buggy endpoints but that's possible of any provider.

chrisdoherty4 (approved):

PR #279, opened 2022-06-07
Redfish streaming multipart uploads

joelrebel:
## What does this PR implement/change/remove?

This change introduces streaming multipart upload for Redfish,
to ensure the firmware file being uploaded is not buffered in memory.

The change enables clients to run multiple updates simultaneously
without consuming as much memory as the firmware files being uploaded.

before...
```
(pprof) top
Showing nodes accounting for 195.35MB, 99.49% of 196.35MB total
Dropped 8 nodes (cum <= 0.98MB)
Showing top 10 nodes out of 22
      flat  flat%   sum%        cum   cum%
  193.34MB 98.47% 98.47%   193.34MB 98.47%  bytes.makeSlice
       2MB  1.02% 99.49%        2MB  1.02%  runtime.allocm
         0     0% 99.49%   193.34MB 98.47%  bytes.(*Buffer).Write
         0     0% 99.49%   193.34MB 98.47%  bytes.(*Buffer).grow
         0     0% 99.49%   193.34MB 98.47%  github.com/bmc-toolbox/bmclib.(*Client).FirmwareInstall
         0     0% 99.49%   193.34MB 98.47%  github.com/bmc-toolbox/bmclib/bmc.FirmwareInstallFromInterfaces
         0     0% 99.49%   193.34MB 98.47%  github.com/bmc-toolbox/bmclib/bmc.firmwareInstall
         0     0% 99.49%   193.34MB 98.47%  github.com/bmc-toolbox/bmclib/providers/redfish.(*Conn).FirmwareInstall
         0     0% 99.49%   193.34MB 98.47%  github.com/bmc-toolbox/bmclib/providers/redfish.(*Conn).runRequestWithMultipartPayload
         0     0% 99.49%   193.34MB 98.47%  io.Copy (inline)
```
after...
```
(pprof) top
Showing nodes accounting for 3072.96kB, 100% of 3072.96kB total
Showing top 10 nodes out of 37
      flat  flat%   sum%        cum   cum%
  512.56kB 16.68% 16.68%   512.56kB 16.68%  runtime.allocm
  512.20kB 16.67% 33.35%   512.20kB 16.67%  runtime.malg
  512.07kB 16.66% 50.01%   512.07kB 16.66%  net/http.ReadResponse
  512.06kB 16.66% 66.68%   512.06kB 16.66%  crypto/hmac.New
  512.05kB 16.66% 83.34%   512.05kB 16.66%  regexp/syntax.(*parser).newRegexp
[...]

PR #345, opened 2023-08-16
Redfish stream multipart uploads

joelrebel:

The Redfish provider currently reads the firmware files into memory before a multipart upload,
this results in high memory consumption for the client and can be problematic for cases
where firmware installs have to be performed simultaneously.

This change was part of bmclib v1 https://github.com/bmc-toolbox/bmclib/pull/279 
but was somehow missed from being included in v2.

Tested with Dell R6515's, R7640.

heap profile - with streaming uploads
```
(pprof) top 20
Showing nodes accounting for 2059.99kB, 100% of 2059.99kB total
Showing top 20 nodes out of 29
      flat  flat%   sum%        cum   cum%
  520.04kB 25.24% 25.24%  1034.42kB 50.21%  encoding/json.(*Decoder).refill
  514.38kB 24.97% 50.21%   514.38kB 24.97%  compress/flate.NewReader
  513.56kB 24.93% 75.15%   513.56kB 24.93%  regexp/syntax.init
  512.01kB 24.85%   100%   512.01kB 24.85%  sync.(*Map).Swap
         0     0%   100%   514.38kB 24.97%  compress/gzip.(*Reader).Reset
         0     0%   100%   514.38kB 24.97%  compress/gzip.(*Reader).readHeader
         0     0%   100%   514.38kB 24.97%  compress/gzip.NewReader
         0     0%   100%  1546.43kB 75.07%  encoding/json.(*Decoder).Decode
         0     0%   100%  1034.42kB 50.21%  encoding/json.(*Decoder).readValue
         0     0%   100%   512.01kB 24.85%  encoding/json.(*decodeState).object
         0     0%   100%   512.01kB 24.85%  encoding/json.(*decodeState).unmarshal
         0     0%   100%   512.01kB 24.85%  encoding/json.(*decodeState).value
         0     0%   100%   512.01kB 24.85%  encoding/json.Unmarshal
         0     0%   100%   512.01kB 24.85%  encoding/json.cachedTypeFields
         0     0%   100%   512.01kB 24.85%  encoding/json.typeEncoder
         0     0%   100%   512.01kB 24.85%  encoding/json.typeFields
         0     0%   100%  1032.05kB 50.10%  github.com/stmcginnis/gofish/common.(*Entity).Get
         0     0%   100%  1032.05kB 50.10%  github.com/stmcginnis/gofish/common.CollectCollection.func1
         0     0%   100%   514.38kB 24.97%  github.com/stmcginnis/gofish/common.CollectList
         0     0%   100%   514.38kB 24.97%  github.com/stmcginnis/gofish/common.GetCollection
```

heap profile - without streaming uploads
```
(pprof) top 30
Showing nodes accounting for 240MB, 99.58% of 241MB total
Dropped 16 nodes (cum <= 1.21MB)
      flat  flat%   sum%        cum   cum%
     240MB 99.58% 99.58%      240MB 99.58%  bytes.growSlice
         0     0% 99.58%      240MB 99.58%  bytes.(*Buffer).Write
         0     0% 99.58%      240MB 99.58%  bytes.(*Buffer).grow
         0     0% 99.58%   240.50MB 99.79%  github.com/bmc-toolbox/bmclib/v2.(*Client).FirmwareInstall
         0     0% 99.58%   240.50MB 99.79%  github.com/bmc-toolbox/bmclib/v2/bmc.FirmwareInstallFromInterfaces
         0     0% 99.58%   240.50MB 99.79%  github.com/bmc-toolbox/bmclib/v2/bmc.firmwareInstall
         0     0% 99.58%   240.50MB 99.79%  github.com/bmc-toolbox/bmclib/v2/providers/redfish.(*Conn).FirmwareInstall
         0     0% 99.58%   240.50MB 99.79%  github.com/bmc-toolbox/bmclib/v2/providers/redfish.(*Conn).runRequestWithMultipartPayload
         0     0% 99.58%      240MB 99.58%  io.Copy (inline)
         0     0% 99.58%      240MB 99.58%  io.copyBuffer
         0     0% 99.58%   240.50MB 99.79%  main.main
         0     0% 99.58%      240MB 99.58%  mime/multipart.(*part).Write
         0     0% 99.58%      241MB   100%  runtime.main
```

### Checklist
- [X] Tests added

```
- Redfish streaming multipart uploads
```

### DIFF
--- a/providers/redfish/firmware.go
+++ b/providers/redfish/firmware.go
@@ -93,7 +93,7 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 	// override the gofish HTTP client timeout,
 	// since the context timeout is set at Open() and is at a lower value than required for this operation.
 	//
-	// record the http client time to be restored
+	// record the http client timeout to be restored
 	httpClientTimeout := c.redfishwrapper.HttpClientTimeout()
 	defer func() {
 		c.redfishwrapper.SetHttpClientTimeout(httpClientTimeout)
@@ -101,9 +101,9 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 
 	c.redfishwrapper.SetHttpClientTimeout(time.Until(ctxDeadline))
 
-	payload := map[string]io.Reader{
-		"UpdateParameters": bytes.NewReader(updateParameters),
-		"UpdateFile":       reader,
+	payload := []map[string]io.Reader{
+		{"UpdateParameters": bytes.NewReader(updateParameters)},
+		{"UpdateFile": reader},
 	}
 
 	resp, err := c.runRequestWithMultipartPayload(http.MethodPost, "/redfish/v1/UpdateService/MultipartUpload", payload)
@@ -197,6 +197,71 @@ func (c *Conn) firmwareUpdateCompatible(ctx context.Context) (err error) {
 	return nil
 }
 
+// pipeReaderFakeSeeker wraps the io.PipeReader and implements the io.Seeker interface
+// to meet the API requirements for the Gofish client https://github.com/stmcginnis/gofish/blob/46b1b33645ed1802727dc4df28f5d3c3da722b15/client.go#L434
+//
+// The Gofish method linked does not currently perform seeks and so a PR will be suggested
+// to change the method signature to accept an io.Reader instead.
+type pipeReaderFakeSeeker struct {
+	*io.PipeReader
+}
+
+// Seek impelements the io.Seeker interface only to panic if called
+func (p pipeReaderFakeSeeker) Seek(offset int64, whence int) (int64, error) {
+	panic("Seek() not implemented for fake pipe reader seeker.")
+}
+
+// multipartPayloadSize prepares a temporary multipart form to determine the form size
+func multipartPayloadSize(payload []map[string]io.Reader) (int64, *bytes.Buffer, error) {
+	body := &bytes.Buffer{}
+	form := multipart.NewWriter(body)
+
+	var size int64
+	var err error
+	for idx, elem := range payload {
+		for key, reader := range elem {
+			var part io.Writer
+			if file, ok := reader.(*os.File); ok {
+				// Add update file fields
+				if _, err = form.CreateFormFile(key, filepath.Base(file.Name())); err != nil {
+					return 0, body, err
+				}
+
+				// determine file size
+				finfo, err := file.Stat()
+				if err != nil {
+					return 0, body, err
+				}
+
+				size = finfo.Size()
+
+			} else {
+				// Add other fields
+				if part, err = updateParametersFormField(key, form); err != nil {
+					return 0, body, err
+				}
+
+				buf := bytes.Buffer{}
+				teeReader := io.TeeReader(reader, &buf)
+
+				if _, err = io.Copy(part, teeReader); err != nil {
+					return 0, body, err
+				}
+
+				// place it back so its available to be read again
+				payload[idx][key] = bytes.NewReader(buf.Bytes())
+			}
+		}
+	}
+
+	err = form.Close()
+	if err != nil {
+		return 0, body, err
+	}
+
+	return int64(body.Len()) + size, body, nil
+}
+
 // runRequestWithMultipartPayload is a copy of https://github.com/stmcginnis/gofish/blob/main/client.go#L349
 // with a change to add the UpdateParameters multipart form field with a json content type header
 // the resulting form ends up in this format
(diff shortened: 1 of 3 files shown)


providers/redfish/firmware.go
```diff
@@ -197,6 +197,71 @@ func (c *Conn) firmwareUpdateCompatible(ctx context.Context) (err error) {
...
+//
+// The Gofish method linked does not currently perform seeks and so a PR will be suggested
+// to change the method signature to accept an io.Reader instead.
+type pipeReaderFakeSeeker struct {
+	*io.PipeReader
+}
+
+// Seek impelements the io.Seeker interface only to panic if called
+func (p pipeReaderFakeSeeker) Seek(offset int64, whence int) (int64, error) {
+	panic("Seek() not implemented for fake pipe reader seeker.")
```
mmlb, line 211: small nit: wouldn't returning an error be better than panic'ing?

mattcburns (approved):

providers/redfish/firmware.go
```diff
@@ -197,6 +197,71 @@ func (c *Conn) firmwareUpdateCompatible(ctx context.Context) (err error) {
...
+
+				// determine file size
+				finfo, err := file.Stat()
+				if err != nil {
+					return 0, body, err
+				}
+
+				size = finfo.Size()
+
+			} else {
```
mmlb, line 238: Any clue what other kinds of readers might show up here? It might be nice to see if the implementation of reader might have a .Len method so we can avoid unnecessary reads.

```go
type Lenner interface {
	Len() int
}
...

if file, ok := reader.(*os.File); ok {
...
} else if lenner, ok := reader.(Lenner); ok {
...
} else {
// read the file to get size
}
```

providers/redfish/firmware.go
```diff
@@ -197,6 +197,71 @@ func (c *Conn) firmwareUpdateCompatible(ctx context.Context) (err error) {
...
+	*io.PipeReader
+}
+
+// Seek impelements the io.Seeker interface only to panic if called
+func (p pipeReaderFakeSeeker) Seek(offset int64, whence int) (int64, error) {
+	panic("Seek() not implemented for fake pipe reader seeker.")
+}
+
+// multipartPayloadSize prepares a temporary multipart form to determine the form size
+func multipartPayloadSize(payload []map[string]io.Reader) (int64, *bytes.Buffer, error) {
```
mmlb, line 215: this func seems rather wasteful, you go through the effort of creating a form only to throw it away to then create another one later in the caller of this func with almost duplicate code... why not just return the form + size from this func?

joelrebel: The `multipartPayloadSize()` function copies the contents of the multipart form into a `bytes.Buffer{}` to determine the actual form size. 
The`runRequestWithMultipartPayload()` uses a `io.PipeWriter` which means the `io.Copy()` blocks until whenever the http request is actually being sent - to pipe the update file contents. 

So I don't see a trivial way to determine the Content-Length without building the multipart form once before actually sending it, I'm open to any ideas thought.

providers/redfish/firmware.go
joelrebel: Hmm, this might need some rethinking - since we also require the file name here,
```go
  if file, ok := reader.(*os.File); ok {
    // Add update file fields
    if _, err = form.CreateFormFile(key, filepath.Base(file.Name())); err != nil {
      return 0, body, err
    }
    ...
  }  
```

providers/redfish/firmware.go
joelrebel: thanks, fixed in [6cf0168](https://github.com/bmc-toolbox/bmclib/pull/345/commits/6cf01686c513a1d94b75b69bed6e7689fd26e5ce)

providers/redfish/firmware.go
mmlb: Aw yeah I see that now. Nope I can't really think of another option w/o checking interfaces for Len and/or Size and maybe avoid the read. :(

joelrebel:
@mmlb whenever you're back, PTAL :)

ofaurax (dismissed):
Waiting for @mmlb but I give my approval since it's needed for #346

providers/redfish/firmware.go
```diff
@@ -197,6 +197,71 @@ func (c *Conn) firmwareUpdateCompatible(ctx context.Context) (err error) {
...
+					return 0, body, err
+				}
+
+				// determine file size
+				finfo, err := file.Stat()
+				if err != nil {
+					return 0, body, err
+				}
+
+				size = finfo.Size()
```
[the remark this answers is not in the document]
mmlb: gotcha, is that described somewhere in this repo (docs maybe)? Either way it would be nice to have a check in this loop maybe to make it obvious to code readers.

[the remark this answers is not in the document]
joelrebel: I took a slightly different route and simplified the implementation - [059b754](https://github.com/bmc-toolbox/bmclib/pull/345/commits/059b754ae098c55eb3b2466e5aaf849e131ea974)

Instead of expecting a payload `[]map[string]io.Reader` its now the `multipartPayload` struct type,
the loop is also unrolled and its simpler to follow the form payload being built.

providers/redfish/firmware.go
```diff
@@ -127,6 +134,11 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 	return taskID, nil
 }
 
+type multipartPayload struct {
+	updateParameters io.Reader
```
mmlb, line 138: So I went looking up the calls stack and now I'm wondering, why does this need to be an io.Reader at all? Its always going to be a `[]byte` from `json.Marshal` in https://github.com/bmc-toolbox/bmclib/blob/a1b87e2ff47f5c05022a6cb8fddc6811e13d9a89/providers/redfish/firmware.go#L79-L87. I don't see why we want to make it a Reader and have to have all the complications of tee'ing and resetting. If its kept as a []byte then down below in multipartPayloadSize p
it'd be something like:

```go
if _, err = part.Write(payload.updateParameters); err != nil {
		return 0, body, err
}
```
right?

mmlb (commented):
Agree that this looks much better! I think if we drop io.Reader for the updateParameters it will similarly make it simpler too.

providers/redfish/firmware.go
joelrebel: thanks for feedback there, [08172f2](https://github.com/bmc-toolbox/bmclib/pull/345/commits/08172f2d91032140fb93567da8046fbcf5b76b0a)

mmlb (approved):
lgtm

PR #346, opened 2023-09-04
Redfish provider to support unstructured http push uploads [WIP]

joelrebel:
### Checklist
- [ ] Tests added

```
```


ofaurax:
@joelrebel please review, it seems you can't be the reviewer of your own PR, but I need a look to my code...

providers/redfish/firmware.go
```diff
@@ -120,59 +124,82 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
 
 	// The response contains a location header pointing to the task URI
```
joelrebel, line 125: lets move this taskID parsing section into a method called - `TaskIDFromLocationURI()` ?

providers/redfish/firmware.go
```diff
@@ -120,59 +124,82 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
...
+
+func (c *Conn) unstructuredHttpUpload(ctx context.Context, url, applyAt string, update io.Reader) (*http.Response, error) {
+
+	if url == "" {
+		return nil, fmt.Errorf("unable to execute request, no target provided")
 	}
 
-	if task == nil {
-		return state, errors.New("failed to lookup task status for task ID: " + taskID)
+	b, _ := io.ReadAll(update)
```
joelrebel, line 176: Can we try to create a goroutine that will io.Copy the contents so we don't have to buffer all the contents here?
See the `runRequestWithMultipartPayload()` method as an example

providers/redfish/tasks.go
```diff
@@ -52,27 +36,6 @@ func (c *Conn) GetFirmwareInstallTaskQueued(ctx context.Context, component strin
 	return task, nil
 }
 
-// purgeQueuedFirmwareInstallTask removes any existing queued firmware install task for the given component slug
-func (c *Conn) purgeQueuedFirmwareInstallTask(ctx context.Context, component string) error {
```
joelrebel: This method is required - https://github.com/bmc-toolbox/bmclib/blob/f096534cec29f6fc4c415a698bd883bdcf5b4894/providers/redfish/firmware.go#L65

providers/redfish/firmware.go
```diff
@@ -409,6 +417,24 @@ func (c *Conn) FirmwareInstallStatus(ctx context.Context, installVersion, compon
 	switch {
 	case strings.Contains(vendor, constants.Dell):
 		task, err = c.dellJobAsRedfishTask(taskID)
+	case strings.Contains(vendor, constants.Packet) || strings.Contains(vendor, constants.Equinix):
```
joelrebel, line 425: lets move this into `tasks.go` and make it more generic method that returns a `*gofish.Task` object,  it can replace the `openbmcGetTask()` method.  
We also don't need to check the vendor, and can just have a default switch case - since this is standard redfish

providers/redfish/firmware.go
```diff
@@ -409,6 +417,24 @@ func (c *Conn) FirmwareInstallStatus(ctx context.Context, installVersion, compon
...
+
+		resp, _ := c.redfishwrapper.Get("/redfish/v1/TaskService/Tasks/" + taskID)
+		if resp.StatusCode != 200 {
+			err = errors.Wrap(
+				bmclibErrs.ErrFirmwareInstall,
```
joelrebel, line 430: This should be a `ErrFirmwareInstallStatus` error instead
 https://github.com/bmc-toolbox/bmclib/blob/redfish-firmware-installs/errors/errors.go#L67

providers/redfish/firmware.go
```diff
@@ -413,6 +413,11 @@ func (c *Conn) FirmwareInstallStatus(ctx context.Context, installVersion, compon
 		return state, errors.Wrap(err, "unable to determine device vendor, model attributes")
 	}
 
+	// component is not used, we hack it for tests
```
joelrebel, line 416: Hardcoding this is not ideal, I think we can make the Task methods more generic, since OpenBMC follows standard Redfish

providers/redfish/firmware_test.go
```diff
@@ -112,17 +112,17 @@ func TestFirmwareInstall(t *testing.T) {
...
-			constants.FirmwareApplyOnReset,
-			false,
-			true,
-			fh,
-			"467696020275",
-			bmclibErrs.ErrFirmwareInstall,
-			"task for BIOS firmware install present",
-			"task ID exists",
-		},
+		//{
```
joelrebel, line 115: This test is most likely failing because of this section being commented out - https://github.com/bmc-toolbox/bmclib/blob/f096534cec29f6fc4c415a698bd883bdcf5b4894/providers/redfish/firmware.go#L65

providers/redfish/firmware.go
ofaurax: Ok for the default, but we still need to check the vendor, as Dell doesn't use the default

providers/redfish/firmware.go
ofaurax: I added a TODO because I don't really understand this for now. I don't know if I really should use a `PipeReader` or if it's related to the multipart. As this is a deprecated update mechanism, we can leave it like that, waiting for it to be removed or improved

providers/redfish/firmware.go
ofaurax: OK, Redfish Task is now the default

providers/redfish/firmware.go
ofaurax: Done

providers/redfish/firmware.go
```diff
@@ -139,52 +160,68 @@ type multipartPayload struct {
...
+	case updateService.MultipartHTTPPushURI != "":
+		return multipartHttpUpload, updateService.MultipartHTTPPushURI, nil
+	case updateService.HTTPPushURI != "":
+		return unstructuredHttpPush, updateService.HTTPPushURI, nil
 	}
 
+	return "", "", errors.Wrap(bmclibErrs.ErrRedfishUpdateService, "unsupported update method")
 }
 
 // firmwareUpdateCompatible retuns an error if the firmware update process for the BMC is not supported
```
joelrebel, line 227: This method can be removed, since its not in use

constants/constants.go
```diff
@@ -24,6 +24,9 @@ const (
 	Quanta = "Quanta"
 	// Quanta is the contant to identify Intel hardware
 	Intel = "Intel"
+	// OpenBMC build of Equinix, formerly Packet
+	Packet = "Packet"
+	Equinix = "Equinix"
```
joelrebel, line 29: This constant can be removed since its not in use

providers/redfish/firmware_test.go
```diff
@@ -233,3 +233,66 @@ func TestFirmwareUpdateCompatible(t *testing.T) {
 		t.Fatal(err)
 	}
```
joelrebel, line 234: This `TestFirmwareUpdateCompatible` can be removed

providers/redfish/firmware.go
```diff
@@ -127,8 +130,26 @@ func (c *Conn) FirmwareInstall(ctx context.Context, component, applyAt string, f
...
+
+func TaskIDFromLocationURI(uri string) (taskID string, err error) {
+
+	if strings.Contains(uri, "JID_") {
+		taskID = strings.Split(uri, "JID_")[1]
+	} else if strings.Contains(uri, "/Monitor") {
+		// OpenBMC returns a monitor URL in Location
+		// Location: /redfish/v1/TaskService/Tasks/12/Monitor
+		splits := strings.Split(uri, "/")
+		taskID = splits[5]
```
joelrebel, line 148: Its better to use regexp here or check the len of splits to make sure this doesn't panic,
```
splits := strings.Split(uri, "/")
if len(splits) == 6 {
    taskID = splits[5]
} else {
    return "", error...    
}
```

providers/redfish/firmware.go
```diff
@@ -139,52 +160,68 @@ type multipartPayload struct {
...
+		updateFile: update,
 	}
 
-	if err != nil {
-		return state, err
+	return c.runRequestWithMultipartPayload(url, payload)
+}
+
+func (c *Conn) unstructuredHttpUpload(ctx context.Context, url, applyAt string, update io.Reader) (*http.Response, error) {
+
```
joelrebel, line 192: purge empty line

providers/redfish/tasks.go
```diff
@@ -1,173 +1,115 @@
...
-		return "JID_" + id
-	}
-
-	return id
-}
-
-func (c *Conn) getDellFirmwareInstallTaskScheduled(slug string) (*gofishrf.Task, error) {
-	// get tasks by state
-	tasks, err := c.dellJobs("scheduled")
+func (c *Conn) activeTask(ctx context.Context) (*gofishrf.Task, error) {
```
joelrebel, line 17: remove unused method

joelrebel (commented):
Thanks for the work here, left a few minor suggestions

providers/redfish/tasks.go
ofaurax: Used at line 46

providers/redfish/firmware.go
ofaurax: Totally true

ofaurax (approved):

ofaurax (approved):

PR #347, opened 2023-09-11
Floppy upload and unmount

joelrebel:

Defines and implements the floppy upload and unmount feature for Supermicro X11, X12 BMCs.
This enables uploading a floppy image onto the BMC, which can then be used to boot off the image.

This differs from mounting remote images as virtual media, which can timeout over time.

### Checklist
- [ ] Tests added

-  Supermicro

- X11, X12

### DIFF
--- a/bmc/floppy.go
+++ b/bmc/floppy.go
@@ -10,19 +10,19 @@ import (
 	"github.com/pkg/errors"
 )
 
-// FloppyImageUploader defines methods to upload a floppy image
-type FloppyImageUploader interface {
-	UploadFloppyImage(ctx context.Context, image io.Reader) (err error)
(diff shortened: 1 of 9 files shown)


DoctorVin (approved):
Seems OK. I find the usage of `Upload` vs. `Unmount` jarring and I'd prefer those be more consistent, but as is this is functional.

joelrebel:
> Seems OK. I find the usage of `Upload` vs. `Unmount` jarring and I'd prefer those be more consistent, but as is this is functional.

You're completely right - I was on fence about the naming, updated now.

joelrebel:
force pushed to rebased on main

DoctorVin (approved):

PR #348, opened 2023-09-19
Fix copying of request body:

jacobweinstock:

The request body was being dropped after the io.Copy causing issues with further reading of it. Add a PingMethod for the Open call. This way we send something for the client to interpret instead of nothing. Update tests.

### Checklist
- [x] Tests added

```
```


joelrebel (commented):
Does the example have to be updated to handle the ping request?

joelrebel (approved):

PR #352, opened 2023-09-20
providers/redfish/tasks: returns TaskNotFound on 404

ofaurax:
### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/providers/redfish/tasks.go
+++ b/providers/redfish/tasks.go
@@ -46,7 +46,7 @@ func (c *Conn) activeTask(ctx context.Context) (*gofishrf.Task, error) {
 
 	err = json.Unmarshal(data, &status)
 	if err != nil {
-		fmt.Println(err)
+		return nil, err
 	}
 
 	// For each task, check if it's running
@@ -122,7 +122,6 @@ func (c *Conn) purgeQueuedFirmwareInstallTask(ctx context.Context, component str
 
 // GetTask returns the current Task fir the given TaskID
 func (c *Conn) GetTask(taskID string) (task *gofishrf.Task, err error) {
-
 	resp, err := c.redfishwrapper.Get("/redfish/v1/TaskService/Tasks/" + taskID)
 	if err != nil {
 		if strings.HasPrefix(err.Error(), "404") {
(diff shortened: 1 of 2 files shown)


providers/redfish/tasks.go
```diff
@@ -126,14 +125,14 @@ func (c *Conn) GetTask(taskID string) (task *gofishrf.Task, err error) {
 
 	resp, err := c.redfishwrapper.Get("/redfish/v1/TaskService/Tasks/" + taskID)
 	if err != nil {
+		if err.Error()[0:3] == "404" {
```
joelrebel, line 128:
```suggestion
		if strings.Contains(err.Error(), "404") {
```

joelrebel (changes requested):

joelrebel:
@ofaurax can you also include a few more fixes - I missed this in the previous PR,

return error instead of fmt.Println,

https://github.com/bmc-toolbox/bmclib/blob/1f97b355960e5ca00f6454317aa89cf8996c776b/providers/redfish/tasks.go#L50

https://github.com/bmc-toolbox/bmclib/blob/1f97b355960e5ca00f6454317aa89cf8996c776b/providers/redfish/tasks.go#L152

remove empty line,
https://github.com/bmc-toolbox/bmclib/blob/1f97b355960e5ca00f6454317aa89cf8996c776b/providers/redfish/tasks.go#L126

else block can be removed,
https://github.com/bmc-toolbox/bmclib/blob/1f97b355960e5ca00f6454317aa89cf8996c776b/providers/redfish/tasks.go#L153
```go
err = json.Unmarshal(data, &status)
	if err != nil {
		return nil, err
	}
	
	task = &gofishrf.Task{
		TaskState:  gofishrf.TaskState(status.TaskState),
		TaskStatus: gofishcommon.Health(status.TaskStatus),
	}

	return task, err
```

providers/redfish/tasks.go
ofaurax: I think we should test the beginning of the string so I'll use `strings.HasPrefix`

joelrebel (approved):

PR #354, opened 2023-09-25
Client opts fixes

joelrebel:

- Exposes the Redfish Etag match header disable parameter added in https://github.com/bmc-toolbox/bmclib/pull/353
- Minor refactor - move provider registration into individual methods

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -165,17 +165,15 @@ func (c *Client) registerIPMIProvider() error {
 }
 
 // register ASRR vendorapi provider
-func (c *Client) registerASRRProvider() error {
+func (c *Client) registerASRRProvider() {
 	asrHttpClient := *c.httpClient
 	asrHttpClient.Transport = c.httpClient.Transport.(*http.Transport).Clone()
 	driverAsrockrack := asrockrack.NewWithOptions(c.Auth.Host+":"+c.providerConfig.asrock.Port, c.Auth.User, c.Auth.Pass, c.Logger, asrockrack.WithHTTPClient(&asrHttpClient))
 	c.Registry.Register(asrockrack.ProviderName, asrockrack.ProviderProtocol, asrockrack.Features, nil, driverAsrockrack)
-
-	return nil
 }
 
 // register gofish provider
-func (c *Client) registerGofishProvider() error {
+func (c *Client) registerGofishProvider() {
 	gfHttpClient := *c.httpClient
 	gfHttpClient.Transport = c.httpClient.Transport.(*http.Transport).Clone()
 	gofishOpts := []redfish.Option{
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -145,50 +145,56 @@ func (c *Client) registerRPCProvider() error {
...
 	}
 
-	// register ASRR vendorapi provider
+	c.Registry.Register(ipmitool.ProviderName, ipmitool.ProviderProtocol, ipmitool.Features, nil, driverIpmitool)
+
+	return nil
+}
+
+// register ASRR vendorapi provider
+func (c *Client) registerASRRProvider() error {
```
jacobweinstock, line 168: this func will never return an error. i would suggest not having any return value.

client.go
```diff
@@ -145,50 +145,56 @@ func (c *Client) registerRPCProvider() error {
...
 	asrHttpClient.Transport = c.httpClient.Transport.(*http.Transport).Clone()
 	driverAsrockrack := asrockrack.NewWithOptions(c.Auth.Host+":"+c.providerConfig.asrock.Port, c.Auth.User, c.Auth.Pass, c.Logger, asrockrack.WithHTTPClient(&asrHttpClient))
 	c.Registry.Register(asrockrack.ProviderName, asrockrack.ProviderProtocol, asrockrack.Features, nil, driverAsrockrack)
 
-	// register gofish provider
+	return nil
+}
+
+// register gofish provider
+func (c *Client) registerGofishProvider() error {
```
jacobweinstock, line 178: this func will never return an error. i would suggest not having any return value.

client.go
```diff
@@ -145,50 +145,56 @@ func (c *Client) registerRPCProvider() error {
...
+
 	driverGoFish := redfish.New(c.Auth.Host, c.Auth.User, c.Auth.Pass, c.Logger, gofishOpts...)
 	c.Registry.Register(redfish.ProviderName, redfish.ProviderProtocol, redfish.Features, nil, driverGoFish)
 
-	// register Intel AMT provider
+	return nil
+}
+
+// register Intel AMT provider
+func (c *Client) registerIntelAMTProvider() error {
```
jacobweinstock, line 196: this func will never return an error. i would suggest not having any return value.

client.go
```diff
@@ -197,7 +203,11 @@ func (c *Client) registerProviders() {
 	driverAMT := intelamt.New(c.Auth.Host, c.Auth.User, c.Auth.Pass, iamtOpts...)
 	c.Registry.Register(intelamt.ProviderName, intelamt.ProviderProtocol, intelamt.Features, nil, driverAMT)
 
-	// register Dell gofish provider
+	return nil
+}
+
+// register Dell gofish provider
+func (c *Client) registerDellProvider() error {
```
jacobweinstock, line 210: this func will never return an error. i would suggest not having any return value.

client.go
```diff
@@ -209,11 +219,64 @@ func (c *Client) registerProviders() {
 	driverGoFishDell := dell.New(c.Auth.Host, c.Auth.User, c.Auth.Pass, c.Logger, dellGofishOpts...)
 	c.Registry.Register(dell.ProviderName, redfish.ProviderProtocol, dell.Features, nil, driverGoFishDell)
 
-	// register supermicro vendorapi provider
+	return nil
+}
+
+// register supermicro vendorapi provider
+func (c *Client) registerSupermicroProvider() error {
```
jacobweinstock, line 226: this func will never return an error. i would suggest not having any return value.

jacobweinstock (changes requested):

client.go
joelrebel: That was mainly keeping the pattern from the above function since I was looking into if these could be turned into an interface method which providers implement to register themselves so this can be moved into each provider.

I can remove the extra error returns for now.

jacobweinstock (approved):

PR #357, opened 2023-10-25
e3c256d4i support

ofaurax:
By default, update is broken on e3c256d4i. This PR fixes it. Compatibility with e3c246d4i needs to be tested

### DIFF
--- a/providers/asrockrack/firmware.go
+++ b/providers/asrockrack/firmware.go
@@ -63,7 +63,16 @@ func (a *ASRockRack) FirmwareInstallStatus(ctx context.Context, installVersion, 
 func (a *ASRockRack) firmwareInstallBMC(ctx context.Context, reader io.Reader, fileSize int64) error {
 	var err error
 
+	// 0. take the model so that we use a different endpoint on E3C256D4ID-NL
+	device := common.NewDevice()
+	device.Metadata = map[string]string{}
+	err = a.fruAttributes(ctx, &device)
+	if err != nil {
+		return errors.Wrap(err, "failed to get model in step 0/4")
+	}
+
 	// 1. set the device to flash mode - prepares the flash
+	// Beware: this locks some capabilities, e.g. the access to fruAttributes
 	a.log.V(2).WithValues("step", "1/4").Info("set device to flash mode, takes a minute...")
 	err = a.setFlashMode(ctx)
 	if err != nil {
@@ -71,8 +80,13 @@ func (a *ASRockRack) firmwareInstallBMC(ctx context.Context, reader io.Reader, f
 	}
 
 	// 2. upload firmware image file
-	a.log.V(2).WithValues("step", "2/4").Info("upload BMC firmware image")
-	err = a.uploadFirmware(ctx, "api/maintenance/firmware", reader, fileSize)
+	fwEndpoint := "api/maintenance/firmware"
+	// E3C256D4ID-NL calls a different endpoint for firmware upload
+	if strings.EqualFold(device.Model, "E3C256D4ID-NL") {
+		fwEndpoint = "api/maintenance/firmware/firmware"
+	}
+	a.log.V(2).WithValues("step", "2/4").Info("upload BMC firmware image to " + fwEndpoint)
+	err = a.uploadFirmware(ctx, fwEndpoint, reader, fileSize)
 	if err != nil {
 		return errors.Wrap(err, "failed in step 2/4 - upload BMC firmware image")
 	}
@@ -146,6 +160,8 @@ func (a *ASRockRack) firmwareUpdateStatus(ctx context.Context, component string,
 		switch progress.State {
 		case 0:
 			return constants.FirmwareInstallRunning, nil
+		case 1: // "Flashing To be done"
+			return constants.FirmwareInstallQueued, nil
 		case 2:
 			return constants.FirmwareInstallComplete, nil
 		default:
--- a/providers/asrockrack/helpers.go
+++ b/providers/asrockrack/helpers.go
@@ -10,9 +10,11 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"os"
+	"strings"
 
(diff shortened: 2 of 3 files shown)


providers/asrockrack/firmware.go
```diff
@@ -74,7 +74,16 @@ func (a *ASRockRack) firmwareInstallBMC(ctx context.Context, reader io.Reader, f
 	a.log.V(2).WithValues("step", "2/4").Info("upload BMC firmware image")
 	err = a.uploadFirmware(ctx, "api/maintenance/firmware", reader, fileSize)
 	if err != nil {
-		return errors.Wrap(err, "failed in step 2/4 - upload BMC firmware image")
+		// Try the special URL for e3c256d4i
```
joelrebel, line 77: Can we check what the error is - if it contains 404, then we attempt the other endpoint.

providers/asrockrack/firmware.go
```diff
@@ -74,7 +74,16 @@ func (a *ASRockRack) firmwareInstallBMC(ctx context.Context, reader io.Reader, f
 	a.log.V(2).WithValues("step", "2/4").Info("upload BMC firmware image")
 	err = a.uploadFirmware(ctx, "api/maintenance/firmware", reader, fileSize)
```
joelrebel, line 75: Would prefer we identify the device model number and call the appropriate endpoint,

The model number can be obtained through this method
https://github.com/bmc-toolbox/bmclib/blob/153e11d62a71723d41a908ad362ab5e6cfc4cd54/providers/asrockrack/inventory.go#L77

providers/asrockrack/helpers.go
```diff
@@ -181,7 +181,15 @@ func (a *ASRockRack) createUpdateUser(ctx context.Context, account *UserAccount)
 // at this point all logged in sessions are terminated
 // and no logins are permitted
 func (a *ASRockRack) setFlashMode(ctx context.Context) error {
-	_, statusCode, err := a.queryHTTPS(ctx, "api/maintenance/flash", "PUT", nil, nil, 0)
+	// preserve config is needed by e3c256d4i
+	pConfig := &preserveConfig{PreserveConfig: 1}
```
joelrebel, line 185: set this configuration based on device model number

joelrebel (changes requested):
Thanks for this submission, can we also make sure this does not cause a regression on the older `e3c246d4i`

providers/asrockrack/helpers.go
ofaurax: OK, but this adds common as dependency

providers/asrockrack/firmware.go
ofaurax: Done

providers/asrockrack/firmware.go
ofaurax: We use the model now

providers/asrockrack/firmware.go
```diff
@@ -63,16 +63,31 @@ func (a *ASRockRack) FirmwareInstallStatus(ctx context.Context, installVersion,
 func (a *ASRockRack) firmwareInstallBMC(ctx context.Context, reader io.Reader, fileSize int64) error {
 	var err error
 
+	// 0. take the model so that we use a different endpoint on E3C256D4ID-NL
+	device := common.NewDevice()
+	device.Metadata = map[string]string{}
+	err = a.fruAttributes(ctx, &device)
+	if err != nil {
+		a.log.V(2).WithValues("step", "0/4").Info("failed to get model, err: " + err.Error())
+		a.log.V(2).WithValues("step", "0/4").Info("continue with default endpoint")
```
joelrebel, line 72: I suggest this returns an error here - if the model cannot be identified, its unlikely to be supported

providers/asrockrack/firmware.go
```diff
@@ -63,16 +63,31 @@ func (a *ASRockRack) FirmwareInstallStatus(ctx context.Context, installVersion,
...
 	if err != nil {
 		return errors.Wrap(err, "failed in step 1/4 - set device to flash mode")
 	}
 
 	// 2. upload firmware image file
-	a.log.V(2).WithValues("step", "2/4").Info("upload BMC firmware image")
-	err = a.uploadFirmware(ctx, "api/maintenance/firmware", reader, fileSize)
+	fw_endpoint := "api/maintenance/firmware"
+	// E3C256D4ID-NL calls a different endpoint for firmware upload
+	if device.Model == "E3C256D4ID-NL" {
```
joelrebel, line 86: Use `strings.EqualFold(device.Model, "E3C256D4ID-NL")`

providers/asrockrack/helpers.go
```diff
@@ -181,7 +182,23 @@ func (a *ASRockRack) createUpdateUser(ctx context.Context, account *UserAccount)
...
 // and no logins are permitted
 func (a *ASRockRack) setFlashMode(ctx context.Context) error {
-	_, statusCode, err := a.queryHTTPS(ctx, "api/maintenance/flash", "PUT", nil, nil, 0)
+	device := common.NewDevice()
+	device.Metadata = map[string]string{}
+	_ = a.fruAttributes(ctx, &device)
+
+	pConfig := &preserveConfig{}
+	// preserve config is needed by e3c256d4i
+	if device.Model == "E3C256D4ID-NL" {
```
joelrebel, line 191: use `strings.EqualFold()` here

joelrebel (changes requested):

providers/asrockrack/firmware.go
ofaurax: Until now, we used the default endpoint for every model, and now only E3C256D4I needs a special one.
So, if the model is unknown, we still try the default one.

However, the code you mention test the error of `fruAttributes` and it's true that if we fail on getting them, we might have a real problem.

joelrebel (approved):
Thanks for the fixes!

PR #358, opened 2023-11-03
Add Odata ID for MegaRAC/AsRockRack Systems & Managers

ofaurax:
Add Odata ID for Systems & Managers (Chassis was already there)

AsRockRack


joelrebel (approved):
Thanks for the fixes 👍 
Going ahead we'll remove the System, Chassis Odata check since its getting prohibitive to support newer hardware.

PR #360, opened 2023-11-07
providers/asrockrack: when checking component, use upper case

ofaurax:
Check component with upper case version (as slugs are upper case in common lib)

asrockrack


joelrebel (approved):
Thanks for the fix

PR #362, opened 2023-11-07
[2/3] New firmware interfaces

joelrebel:

Note: to be merged after https://github.com/bmc-toolbox/bmclib/pull/363

Defines new firmware interface methods - to enable individual firmware actions instead of a larger abstraction like the current `FirmwareInstall`, `FirmwareInstallStatus` - the abstraction in these current methods is getting to large to manage and work across vendors. 

`FirmwareUpload` - Only upload firmware
`FirmwareInstallUploaded` - Install previously uploaded firmware
`FirmwareTaskStatus` - Retrieve task state along with status information, to deprecate `FirmwareInstallStatus`
`FirmwareInstallSteps` - Returns steps the caller has to follow to install firmware for a device, since each vendor device is a special snowflake.

### Checklist
- [X] Tests added


bmc/firmware.go
```diff
@@ -5,25 +5,27 @@ import (
...
 
-// FirmwareInstaller defines an interface to install firmware updates
+// FirmwareInstaller defines an interface to upload and initiate a firmware install
 type FirmwareInstaller interface {
 	// FirmwareInstall uploads firmware update payload to the BMC returning the task ID
 	//
 	// parameters:
 	// component - the component slug for the component update being installed.
-	// applyAt - one of "Immediate", "OnReset".
+	// operationsApplyTime - one of the OperationApplyTime constants
```
DoctorVin, line 22: Why not make that a type instead of constants of `string`?

bmc/firmware.go
```diff
@@ -165,3 +169,294 @@ func FirmwareInstallStatusFromInterfaces(ctx context.Context, installVersion, co
...
+}
+
+// firmwareInstallerProvider is an internal struct to correlate an implementation/provider and its name
+type firmwareInstallerWithOptionsProvider struct {
+	name string
+	FirmwareInstallerUploaded
+}
+
+// firmwareInstallUploaded uploads and initiates firmware update for the component
+func firmwareInstallUploaded(ctx context.Context, component, uploadTaskID string, generic []firmwareInstallerWithOptionsProvider) (installTaskID string, metadata Metadata, err error) {
```
DoctorVin, line 193: In general I think named returns are a maintenance hazard.

bmc/firmware.go
joelrebel: Yep, that is the intention going ahead in the newer interface, for now we don't want to break the current interface.

bmc/firmware.go
joelrebel: I kinda agree, although I have them there since they are indicative of what the returned string holds.

joelrebel:
rebase on parent branch

PR #366, opened 2023-11-16
Support for Asrockrack e3C256D4ID-NL

ofaurax:
Asrockrack e3C256D4ID-NL

### DIFF
--- a/providers/asrockrack/firmware.go
+++ b/providers/asrockrack/firmware.go
@@ -184,6 +184,13 @@ func (a *ASRockRack) firmwareUpdateStatus(ctx context.Context, component string,
 
 	switch installStatus {
 	case versionStrMatch:
+		if progress == nil {
+			// TODO: we should pass the force parameter to firmwareUpdateStatus,
+			// so that we can know if we expect a version change or not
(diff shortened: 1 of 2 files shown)


providers/asrockrack/firmware.go
```diff
@@ -182,6 +182,13 @@ func (a *ASRockRack) firmwareUpdateStatus(ctx context.Context, component string,
...
 
+	if progress == nil && installStatus == versionStrMatch {
+		// TODO: we should pass the force parameter to firmwareUpdateStatus,
+		// so that we can know if we expect a version change or not
+		a.log.V(3).Info("Nil progress + no version change -> unknown")
+		return constants.FirmwareInstallUnknown, nil
+	}
+
 	switch installStatus {
 	case versionStrMatch:
```
joelrebel, line 193: I think you can move the added conditional `if progress == nil { return constants.FirmwareInstallUnknown} }` under this switch case?

joelrebel (commented):
Thanks for the fix - suggested a minor change

providers/asrockrack/firmware.go
ofaurax: Sure.

joelrebel (approved):

PR #367, opened 2023-11-18
Add GetBootDeviceOverride support for redfish

coffeefreak101:
Adds a method to the client for getting boot device override settings.

The information returned includes:
- If the override is persistent
- If the override mode is UEFI
- The override device

The output was designed to be similar to the parameters of the `BootDeviceSet` method.

### Checklist
- [x] Tests added

Redfish APIs


bmc/boot_device.go
```diff
@@ -14,12 +14,29 @@ type BootDeviceSetter interface {
...
+// bootOverrideProvider is an internal struct to correlate an implementation/provider and its name
+type bootOverrideProvider struct {
+	name          string
+	bootOverrider BootDeviceOverrideGetter
+}
+
+type BootDeviceOverride struct {
+	IsPersistent bool
+	IsEFIBoot    bool
+	Device       string
```
joelrebel, line 37: What can be expected to be returned as `Device` here - does that differ per vendor?

bmc/boot_device.go
```diff
@@ -78,3 +95,67 @@ func SetBootDeviceFromInterfaces(ctx context.Context, timeout time.Duration, boo
...
+
+// GetBootDeviceOverrideFromInterface will get boot device override settings from the first successful
+// call to a BootDeviceOverrideGetter in the array of providers.
+func GetBootDeviceOverrideFromInterface(
+	ctx context.Context,
+	timeout time.Duration,
+	providers []interface{},
+) (*BootDeviceOverride, Metadata, error) {
+	var err error
+	metadata := Metadata{
```
joelrebel, line 135: for consistency use the [`newMetadata()`](https://github.com/bmc-toolbox/bmclib/blob/b0ef181d6ed441ea5bd9e7161e92712633a28811/bmc/bmc.go#L24) helper method here

bmc/boot_device.go
```diff
@@ -78,3 +95,67 @@ func SetBootDeviceFromInterfaces(ctx context.Context, timeout time.Duration, boo
...
+		FailedProviderDetail: make(map[string]string),
+	}
+
+	for _, elem := range providers {
+		provider := &bootOverrideProvider{name: getProviderName(elem)}
+		switch p := elem.(type) {
+		case BootDeviceOverrideGetter:
+			provider.bootOverrider = p
+			override, getErr := getBootDeviceOverride(ctx, timeout, provider, &metadata)
```
joelrebel, line 144: For consistency in these interface identifier methods, I suggest collecting all the implementations of `BootDeviceOverrideGetters` in a slice and returning `getBootDeviceOverride` at the end, similar to https://github.com/bmc-toolbox/bmclib/blob/b0ef181d6ed441ea5bd9e7161e92712633a28811/bmc/boot_device.go#L42 
Unless you have some reason to not follow that pattern here

errors/errors.go
```diff
@@ -117,6 +117,9 @@ var (
 
 	// ErrSystemVendorModel is returned when the system vendor, model attributes could not be identified.
 	ErrSystemVendorModel = errors.New("error identifying system vendor, model attributes")
+
+	// ErrNoSystemsAvailable is returned when the API of the device provides and empty array of systems.
+	ErrNoSystemsAvailable = errors.New("no systems were found on the device")
```
joelrebel, line 122: can this be `ErrRedfishNoSystems = errors.New("redfish: no Systems were found on device")` instead, so its a bit more indicative of being a Redfish error?

internal/redfishwrapper/boot_device.go
```diff
@@ -76,3 +77,65 @@ func (c *Client) SystemBootDeviceSet(ctx context.Context, bootDevice string, set
 
 	return true, nil
 }
+
+// bootTargetToDevice tries to convert the redfish boot target to a bmclib supported device string.
+// if the target is unknown or unsupported, then a string of the target is returned.
+func bootTargetToDevice(target rf.BootSourceOverrideTarget) (device string) {
+	switch target {
```
joelrebel, line 84: Can this switch case be moved into its own method and be reused in this method and the above `SystemBootDeviceSet` method?

bmc/boot_device.go
coffeefreak101: I did notice that implementation. I decided to implement it differently primarily because `setBootDevice` uses a `defer` inside a `for` loop, which can cause multiple deferred executions to build up until the function returns. I know we don't really have that many things to loop over, but thought I would at least try to do the right thing and move the defer.

Also, I thought it would be more simplistic to loop over the slice of providers once instead of twice (once to filter, then again to do the calls).

Do you still recommend I filter and then return `getBootDeviceOverride` for consistency?

internal/redfishwrapper/boot_device.go
coffeefreak101: The switch statement is in it's own function, but not a method. Is that ok? I didn't see a need to make it a method since it doesn't make use of the client.

The `SystemBootDeviceSet` method above is actually doing the inverse of this switch statement (string => `rf.BootSourceOverrideTarget`), but I can write something that works for both scenarios.

bmc/boot_device.go
coffeefreak101: I would definitely prefer to have this be an enum or similar. However, `SetBootDevice` takes the boot device as a string, and I thought this information should match that interface.

```
func (c *Client) SetBootDevice(ctx context.Context, bootDevice string, setPersistent, efiBoot bool) (ok bool, err error) {
```

bmc/boot_device.go
```diff
@@ -14,12 +14,29 @@ type BootDeviceSetter interface {
 	BootDeviceSet(ctx context.Context, bootDevice string, setPersistent, efiBoot bool) (ok bool, err error)
 }
 
+// BootDeviceOverrideGetter gets boot override settings for a machine
+type BootDeviceOverrideGetter interface {
+	BootDeviceOverrideGet(ctx context.Context) (override *BootDeviceOverride, err error)
```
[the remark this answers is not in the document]
coffeefreak101: Done

bmc/boot_device.go
coffeefreak101: Done

errors/errors.go
coffeefreak101: Done

bmc/boot_device.go
joelrebel: I'd like to use constants where we can going ahead (for less ambiguity). 
Lets update the new method to accept a constant, and create an issue/PR to update `SetBootDevice`.

client.go
```diff
@@ -455,6 +455,13 @@ func (c *Client) ReadUsers(ctx context.Context) (users []map[string]string, err
 	return users, err
 }
 
+// GetBootDeviceOverride pass through to library function
+func (c *Client) GetBootDeviceOverride(ctx context.Context) (override bmc.BootDeviceOverride, err error) {
```
joelrebel, line 459: Could you also include the trace span
```
ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "GetBootDeviceOverride")
defer span.End()
```

bmc/boot_device.go
joelrebel: Ah, now that I see the reasoning for the difference, it makes sense to me. As long as the behaviour is consistent and new providers can be added without any changes here - this is fine. Thanks!
At some point we need to figure if these boilerplate like methods can be generated along with tests.

internal/redfishwrapper/boot_device.go
joelrebel: that works, thanks!

client.go
coffeefreak101: Thanks for mentioning that. Done.

joelrebel (approved):

bmc/boot_device.go
coffeefreak101: Done. Mind having another look?

joelrebel (approved):

PR #370, opened 2023-11-27
Supermicro redfish methods

joelrebel:

- Implements the Inventory, PowerStateGet, PowerSet methods for the Supermicro provider using the redfishwrapper package.

### Checklist
- [ ] Tests added

### DIFF
--- a/providers/supermicro/supermicro.go
+++ b/providers/supermicro/supermicro.go
@@ -249,21 +249,6 @@ func (c *Client) bmcQueryor(ctx context.Context) (bmcQueryor, error) {
 	return queryor, nil
 }
 
-func (c *Client) openRedfish(ctx context.Context) error {
-	if c.serviceClient.redfish != nil && c.serviceClient.redfish.SessionActive() == nil {
-		return nil
-	}
-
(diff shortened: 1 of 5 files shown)


providers/supermicro/supermicro.go
```diff
@@ -184,6 +188,33 @@ func (c *Client) Open(ctx context.Context) (err error) {
...
+}
+
+// PowerSet sets the power state of a server
+func (c *Client) PowerSet(ctx context.Context, state string) (ok bool, err error) {
+	if c.serviceClient == nil || c.serviceClient.redfish == nil {
+		return false, errors.Wrap(bmclibErrs.ErrLoginFailed, "client not initialized")
+	}
+
+	return c.serviceClient.redfish.PowerSet(ctx, state)
+}
```
coffeefreak101, line 207: Should the `state` for these methods be a constant with a specific type, something like  `PowerState`?

joelrebel: yep thats the plan, although since this will affect _all_ providers and it changes the interface, it needs to be done separately

DoctorVin (approved):

coffeefreak101 (approved):

PR #373, opened 2023-11-28
Update firmware install status for scheduling state

coffeefreak101:

During my firmware update testing, I encountered a situation where the firmware update status went from "initializing" to "running" for just a second, then to "powercycle-host" then back to "running".

This occurred because the BMC's "scheduling" state gets reported as "running", and the "scheduled" state is reported as "powercycle-host".

This change moves "scheduling" BMC state to the "initializing" firmware update status.


joelrebel:
good catch,
Since theres an ongoing overhaul in these changes - do you mind basing this PR off this branch https://github.com/bmc-toolbox/bmclib/tree/redfish-dell

coffeefreak101:
> good catch, Since theres an ongoing overhaul in this area - do you mind basing this PR off this branch https://github.com/bmc-toolbox/bmclib/tree/redfish-dell

I see you're moving the firmware file in that branch. Looks like your PR has the approvals. I can wait for that to merge first before merging this one.

coffeefreak101:
@joelrebel 

I rebased and fixed this base on the new code.

joelrebel (approved):
Thanks for fixing this up

PR #376, opened 2023-11-30
Get systemeventlog

mattcburns:

This PR adds the ability to retrieve the System Event Log from the target BMC using either IPMI or Redfish. There is an opinionated method that returns the entries as an array of entries in array format and a raw method that returns the raw output from the BMC.

### Checklist
- [X] Tests added

N/A
N/A
N/A
N/A

```
Added GetSystemEventLog and GetSystemEventLogRaw. These methods retrieve the SEL of the target BMC. The former uses an opinionated format, retrieving the ID, Timestamp, Description, Message in an array format. The latter returns the raw output as a string. Both IPMI and Redfish are supported.
```


internal/ipmi/ipmi.go
```diff
@@ -383,3 +383,40 @@ func (i *Ipmi) ClearSystemEventLog(ctx context.Context) (err error) {
...
+		}
+		if line[0] == "ID" {
+			continue
+		}
+		for i := range line {
+			line[i] = strings.TrimSpace(line[i])
+		}
+
+		// ID, Timestamp (date time), Description, Message (message : assertion)
+		entries = append(entries, []string{line[0], fmt.Sprintf("%s %s", line[1], line[2]), line[3], fmt.Sprintf("%s : %s", line[4], line[5])})
```
joelrebel, line 408: since this indexes `line[3], line[4], line[5]` - it would be worth adding a `len()` check before referencing those elem indexes.

internal/ipmi/ipmi.go
```diff
@@ -383,3 +383,40 @@ func (i *Ipmi) ClearSystemEventLog(ctx context.Context) (err error) {
...
+			continue
+		}
+		if line[0] == "ID" {
+			continue
+		}
+		for i := range line {
+			line[i] = strings.TrimSpace(line[i])
+		}
+
+		// ID, Timestamp (date time), Description, Message (message : assertion)
```
joelrebel, line 407: can you include some fixture data for this, along with a test for the parsing (move this scanner block into another method if that makes it easier to test)

joelrebel (changes requested):
Thanks for the work here, left a few comments

joelrebel (approved):
Thanks for the fixes and implementation!

joelrebel (approved):
Thanks for the fixes and implementation!

(creds to GH vscode UI for the dup comment)

PR #381, opened 2023-12-13
Add TerminateSOL method

zevweiss:

A new operation to terminate an active SOL session.

### Checklist
- [x] Tests added

```
A new `TerminateSOL` operation, via the ipmitool provider, can be used to terminate an active SOL session.
```

### DIFF
--- a/bmc/sol.go
+++ b/bmc/sol.go
@@ -1,0 +1,70 @@
+package bmc
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+// SOLDeactivator for deactivating SOL sessions on a BMC.
+type SOLDeactivator interface {
+	DeactivateSOL(ctx context.Context) (err error)
+}
+
+// deactivatorProvider is an internal struct to correlate an implementation/provider and its name
+type deactivatorProvider struct {
+	name           string
+	solDeactivator SOLDeactivator
+}
+
+// deactivateSOL tries all implementations for a successful SOL deactivation
+func deactivateSOL(ctx context.Context, timeout time.Duration, b []deactivatorProvider) (metadata Metadata, err error) {
+	var metadataLocal Metadata
+
+	for _, elem := range b {
+		if elem.solDeactivator == nil {
+			continue
+		}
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -526,6 +526,13 @@ func (c *Client) ResetBMC(ctx context.Context, resetType string) (ok bool, err e
 	return ok, err
 }
 
+// TerminateSOL pass through library function to terminate active SOL sessions
+func (c *Client) TerminateSOL(ctx context.Context) (ok bool, err error) {
```
joelrebel, line 530: Include tracing for the client method
```suggestion
func (c *Client) TerminateSOL(ctx context.Context) (ok bool, err error) {
    ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "TerminateSOL")
	defer span.End()
```

bmc/sol.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+// SOLTerminator for terminating SOL sessions on a BMC.
+type SOLTerminator interface {
```
joelrebel, line 13: I prefer if this was named `SolDeactivator` instead and similarly for its implementations methods, did you have reasons to call it otherwise?

joelrebel (commented):
Thanks for adding this feature, left some feedback on the naming of this method.

bmc/sol.go
zevweiss: Heh -- sure, I'd be fine with `SOLDeactivator`.

client.go
zevweiss: Ack, will do -- thanks.

bmc/sol.go
zevweiss: ...and to clarify, would you prefer that just for this particular interface, or a broader `s/terminate/deactivate/` across the whole patch?

joelrebel: @zevweiss a broader `s/terminate/deactivate/` across the whole patch would be ideal

zevweiss: Ack, done.

joelrebel (approved):

PR #383, opened 2024-02-27
[OpenBMC] Return err if open fails:

jacobweinstock:

Returning the error will ensure the provider is removed from future calls.

### Checklist
- [ ] Tests added

```
```


joelrebel (approved):
Thanks for the fix!

PR #384, opened 2024-03-07
Add support for sending NMI

coffeefreak101:
Support for sending an NMI has been added to ipmi, redfish, redfishwrapper, and all providers that use the redfishwrapper.

### Checklist
- [x] Tests added

```
Adds support for sending an NMI to ipmi, redfish, redfishwrapper, and all providers that use the redfishwrapper.
```

### DIFF
--- a/internal/redfishwrapper/power.go
+++ b/internal/redfishwrapper/power.go
@@ -212,7 +212,6 @@ func (c *Client) SystemForceOff(ctx context.Context) (ok bool, err error) {
 
 		system.DisableEtagMatch(c.disableEtagMatch)
 
-
 		err = system.Reset(rf.ForceOffResetType)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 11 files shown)


internal/redfishwrapper/power.go
```diff
@@ -221,3 +220,19 @@ func (c *Client) SystemForceOff(ctx context.Context) (ok bool, err error) {
 
 	return true, nil
 }
+
+// SendNMI tells the BMC to issue an NMI to the device
+func (c *Client) SendNMI(_ context.Context) error {
```
joelrebel, line 225: before invoking `Systems()`, add in a session check - so we don't end up with nil panics because the redfish session wasn't initialised or has expired.
```
	if err := c.SessionActive(); err != nil {
		return false, errors.Wrap(bmclibErrs.ErrNotAuthenticated, err.Error())
	}
```

joelrebel (commented):
Looks good to me, just one minor suggestion

internal/redfishwrapper/power.go
coffeefreak101: Good catch, done.

joelrebel (approved):
👍

PR #385, opened 2024-04-11
SetBiosConfiguration(ctx, biosConfig) & ResetBiosConfiguration(ctx) for dell

splaspood:

This PR adds support for setting BIOS configuration values via the redfishwrapper on Dell platforms.    Additionally, the ability to perform a 'factory reset' of the BIOS configuration, also for Dell platforms, is included.

### Checklist
- [ ] Tests added

Dell

```
Adds SetBiosConfiguration() and ResetBiosConfiguration() to the dell provider.
```


splaspood:
Pending addition of ResetBiosConfiguration() functionality, and tests.

splaspood:
Added ResetBiosConfiguration() functionality.      Working up some testing now..

client.go
```diff
@@ -556,6 +555,28 @@ func (c *Client) GetBiosConfiguration(ctx context.Context) (biosConfig map[strin
 	return biosConfig, err
 }
 
+func (c *Client) SetBiosConfiguration(ctx context.Context, biosConfig map[string]string) (err error) {
```
coffeefreak101, line 558: Do you think it would be possible to have the map keys of `biosConfig` be some `const` value? Or maybe use a structure with fields for specific settings? Using a map with string keys and values makes it hard for the caller to know what settings are available, or supported.

coffeefreak101 (approved):

splaspood:
> Nice work, mind adding in an example?

There is one, examples/bios exercises GetBiosConfiguration(), SetBiosConfiguration() and ResetBiosConfiguration()

joelrebel (approved):

PR #387, opened 2024-04-17
Smc checks

joelrebel:
- list `x12spo-ntf` as supported.
- Ensure session is closed on a partial `Open()` failure. 

### Checklist
- [ ] Tests added

SMC


DoctorVin (dismissed):
Need to update `golangci-lint`?

joelrebel:
> Need to update `golangci-lint`?

yep, I'd prefer to do that in a follow up PR.

DoctorVin (approved):

PR #388, opened 2024-04-17
Supported check

joelrebel:
Verify hardware support in each Redfish based provider.

This is to ensure each of these hardware vendor specific providers will not proceed to execute methods unless the hardware is identified as supported.

### Checklist
- [ ] Tests added

- Dell
- ASRR

### DIFF
--- a/providers/dell/idrac.go
+++ b/providers/dell/idrac.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"strings"
 
+	bmcliberrs "github.com/bmc-toolbox/bmclib/v2/errors"
 	"github.com/bmc-toolbox/bmclib/v2/internal/httpclient"
 	"github.com/bmc-toolbox/bmclib/v2/internal/redfishwrapper"
 	"github.com/bmc-toolbox/bmclib/v2/providers"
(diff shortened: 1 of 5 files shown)


providers/dell/idrac.go
```diff
@@ -204,21 +205,37 @@ func (c *Conn) Compatible(ctx context.Context) bool {
 
 // PowerStateGet gets the power state of a BMC machine
 func (c *Conn) PowerStateGet(ctx context.Context) (state string, err error) {
+	if err := c.deviceSupported(ctx); err != nil {
```
jacobweinstock, line 208: This provider already has this check in `Open`. I'm not following the value add for putting this in all these individual methods. As far as I can tell `deviceSupported` doesnt do any method specific checks either.

joelrebel: Thanks for pointing that out, you're right the dell, openbmc checks were redundant. 

I noticed dell showing up as a `successfulProvider` on another device - which was odd, but I haven't been able to reproduce the case, and theres no way it should end up in there. Changes reverted.

jacobweinstock (approved):

PR #386, opened 2024-04-11
Upgrade go to 1.22 and dependencies

splaspood:
## What does this PR implement/change/remove?

Upgrade from go 1.18 to 1.22.

## Description for changelog/release notes

```
Upgrade to go 1.22 as well as refresh of deps.
```

PR #390, opened 2024-05-03
Revert to Go 1.18 in go.mod and make policy around this official

jacobweinstock:
We don't seem to have an official strategy for the Go version in go.mod. I believe that, as a library, we should only bump this version if there are dependencies we use that would require us to move to this version ([ref](https://go.dev/doc/modules/gomod-ref#:~:text=The%20minimum%20version%20of%20Go%20required%20by%20the%20current%20module.)). I don't believe that we have any dependencies that warrant a bump. Bumping this version will require user to update to this version. While it is a good practice to build with the latest version of Go, this requirement will be making a decision for users that should be theirs and not ours. We are a library not a binary to be built. In this aspect, we are using the version in go.mod incorrectly, in my opinion, by doing this. This assumes we don't have a known use case where bumping this version if needed. I don't know of any at the moment. If there are any we should bring them up.

Apologies, i know i approved the [PR](https://github.com/bmc-toolbox/bmclib/pull/386) to update this. The affects of this didn't dawn on me then.

### __The other side of the coin__

There can be reasons to bump this version. I think before we bump the version we should have use cases the bump and more carefully consider the affects. From the official Go docs:
_"At go 1.21 or higher:
The go line declares a required minimum version of Go to use with this module.
The go line must be greater than or equal to the go line of all dependencies.
The go command no longer attempts to maintain compatibility with the previous older version of Go.
The go command is more careful about keeping checksums of go.mod files in the go.sum file."_ - [ref](https://go.dev/doc/modules/gomod-ref#:~:text=At%20go%201.21,go.sum%20file.)

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -88,7 +88,6 @@ golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.19.0 h1:q5f1RH2jigJ1MoAWp2KTp3gm5zAGFUTarQZ5U386+4o=
 golang.org/x/sys v0.19.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/term v0.19.0 h1:+ThwsDv+tYfnJFhF4L8jITxu1tdTWRTZpdsWgEgjL6Q=
-golang.org/x/term v0.19.0/go.mod h1:2CuTdWZ7KHSQwUzKva0cbMg6q2DMI3Mmxp+gKJbskEk=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c h1:Hei/4ADfdWqJk1ZMxUNpqntNwaWcugrBjAiHlqqRiVk=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c/go.mod h1:JHkPIbrfpd72SG/EVd6muEfDQjcINNoR0C8j2r3qZ4Q=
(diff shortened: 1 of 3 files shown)


joelrebel:
@jacobweinstock based on the Go release policy, we'd want to be on a release that is not at its EOL?
> Each major Go release is supported until there are two newer major releases
ref: https://go.dev/doc/devel/release#policy

ofaurax:
Note that is the *minimum* version supported.

By setting it to 1.18, we don't prevent people from using latest version.
But by setting it to 1.22 (or oldest not EOL), we just force people to upgrade without technical justification other than "keep your go secure" (which is not this project's mission).

I see several options here: we could stay on the smallest go version of dependencies, or follow the base go version available for some platforms (e.g. min of go version available in oldest supported debian/redhat/ubuntu)

go.sum
```diff
@@ -88,7 +88,6 @@ golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.19.0 h1:q5f1RH2jigJ1MoAWp2KTp3gm5zAGFUTarQZ5U386+4o=
 golang.org/x/sys v0.19.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/term v0.19.0 h1:+ThwsDv+tYfnJFhF4L8jITxu1tdTWRTZpdsWgEgjL6Q=
-golang.org/x/term v0.19.0/go.mod h1:2CuTdWZ7KHSQwUzKva0cbMg6q2DMI3Mmxp+gKJbskEk=
```
ofaurax, line 91: Is this an effect of the downgrading?

chrisdoherty4:
I agree with the OP: libraries should honor the definition of the `go` directive and specify what they need (apps probably should also as of 1.21 which introduces `toolchain`).

> I see several options here: we could stay on the smallest go version of dependencies, or follow the base go version available for some platforms (e.g. min of go version available in oldest supported debian/redhat/ubuntu)

I think the only reason to care about supported platforms is in implementation detail because you wouldn't want to use a language feature > than the oldest supported platform. 

Assuming there aren't any language feature issues I would stick with the definition of the `go` directive and select the oldest version that can be used for the language features leveraged.

joelrebel:
Those are fair points, thanks for raising them. If we can get a rebase on main here then this can be merged.

splaspood (approved):

joelrebel (approved):

PR #392, opened 2024-05-28
FS-1123: Supermicro (SMC) BIOS config Get, Set and Reset support via SUM

splaspood:

BIOS config management support for Supermicro (SMC) platforms implemented as a wrapper over Supermicro Update Manager (SUM).
- GetBiosConfiguration
- SetBiosConfiguration
- SetBiosConfigurationFromFile
- ResetBiosConfiguration

BIOS config management support for Supermicro (SMC) platforms.

```
```


splaspood:
Still need to complete some code here.

mkukiell:
@splaspood is this ready for review? set reviewers in that case.

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+	// ProviderName for the provider implementation
+	ProviderName = "sum"
+	// ProviderProtocol for the provider implementation
+	ProviderProtocol = "sum"
+)
+
+var (
+	// Features implemented by SUM
+	Features = registrar.Features{
+		providers.FeatureGetBiosConfiguration,
```
joelrebel, line 26: These should go into the supermicro provider

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+	biosConfig, err = vcm.StandardConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	return biosConfig, nil
+}
+
+// SetBiosConfiguration set bios configuration
+func (s *Exec) SetBiosConfiguration(ctx context.Context, biosConfig map[string]string) (err error) {
```
joelrebel, line 172: could we have some test cases here that verify the expected command is being executed

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (s *Exec) run(ctx context.Context, command string, additionalArgs ...string) (output string, err error) {
+	// TODO(splaspood) use a tmp file here (as sum supports) to read the password
+	sumArgs := []string{"-i", s.Host, "-u", s.Username, "-p", s.Password, "-c", command}
+	sumArgs = append(sumArgs, additionalArgs...)
+
+	s.Log.V(9).WithValues(
+		"sumArgs",
+		sumArgs,
+	).Info("Calling sum")
```
joelrebel, line 107: suggest using Debug or Trace instead of Info

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+// Close a connection to a BMC
+func (c *Exec) Close(ctx context.Context) (err error) {
+	return nil
+}
+
+func (c *Exec) Name() string {
+	return ProviderName
+}
+
+func (s *Exec) run(ctx context.Context, command string, additionalArgs ...string) (output string, err error) {
```
joelrebel, line 99: add tests

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+func (s *Exec) GetCurrentBiosCfg(ctx context.Context) (output string, err error) {
+	return s.run(ctx, "GetCurrentBiosCfg")
+}
+
+func (s *Exec) LoadDefaultBiosCfg(ctx context.Context) (err error) {
+	_, err = s.run(ctx, "LoadDefaultBiosCfg")
+	return
+}
+
+func (s *Exec) ChangeBiosCfg(ctx context.Context, cfgFile string, reboot bool) (err error) {
```
joelrebel, line 133: add tests

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+			case "Enabled":
+				if err = vcm.SRIOV(true); err != nil {
+					return err
+				}
+			case "Disabled":
+				if err = vcm.SRIOV(false); err != nil {
+					return err
+				}
+			}
+		case strings.HasPrefix(k, "raw:"):
```
joelrebel, line 236: The raw prefix is an ironlib thing, can we instead have method in bmc-toolbox that adds and removes this prefix, and that method is invoked here and in ironlib. This will provide more context into this prefix value.

providers/sum/sum.go
```diff
@@ -0,0 +1,280 @@
...
+	return string(b_out), err
+}
+
+func (s *Exec) GetCurrentBiosCfg(ctx context.Context) (output string, err error) {
+	return s.run(ctx, "GetCurrentBiosCfg")
+}
+
+func (s *Exec) LoadDefaultBiosCfg(ctx context.Context) (err error) {
+	_, err = s.run(ctx, "LoadDefaultBiosCfg")
+	return
```
joelrebel, line 130: use implicit returns through this code base

providers/sum/sum.go
splaspood: I believe this logger uses that V(9) to define what level of 'verbosity' something is logged at?    Correct me if I'm wrong...   but there's no .Debug() available off that as-is.

joelrebel: you're right there, ignore my suggestion

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/go:1-1.21-bullseye
+FROM mcr.microsoft.com/devcontainers/go:1-1.22-bullseye
```
DoctorVin, line 1: why are we using a Microsoft golang container?

internal/executor/executor.go
```diff
@@ -0,0 +1,168 @@
...
+	SetVerbose()
+	GetCmd() string
+	DisableBinCheck()
+	SetStdin(io.Reader)
+	CmdPath() string
+	CheckExecutable() error
+	// for tests
+	SetStdout([]byte)
+	SetStderr([]byte)
+	SetExitCode(int)
```
DoctorVin, line 29: Do these have to be exported?

internal/executor/fake_executor.go
DoctorVin, line 1: Why not use `mockery` or `gomock` here?

DoctorVin (changes requested):
I think hand-making your own mocks is not useful here. Using a test-mock generator will allow you to avoid having to expose test infrastructure in the interface and give you better flexibility.

.devcontainer/Dockerfile
splaspood: I'd guess simply because that's what's suggested in the devcontainer documentation/examples.     My understanding is there's nothing really special about the devcontainer image(s) so we could replace that with anything.

internal/executor/executor.go
splaspood: I'd have to review the code in more detail, but my guess would be these are exported for use in testing.    They are exported here because they were exported in the ironlib code that this was directly copied from.

internal/executor/fake_executor.go
splaspood: Only because this already existed in our code base.    I don't have anything against using these libs...

internal/executor/executor.go
```diff
@@ -0,0 +1,168 @@
...
+	"context"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// Executor interface lets us implement dummy executors for tests
```
jacobweinstock, line 14: this comment says this is for tests but this interface is used the the `sum` package which is not a test package. which one is it, for tests or not?

Also, why is this a package at all? As far as i can tell this is only used in the `sum` package. This should probably just be move there.

internal/sum/sum.go
```diff
@@ -0,0 +1,279 @@
...
+	"strings"
+
+	ex "github.com/bmc-toolbox/bmclib/v2/internal/executor"
+
+	"github.com/bmc-toolbox/common"
+	"github.com/bmc-toolbox/common/config"
+	"github.com/go-logr/logr"
+)
+
+// Mvcli is a mvcli command executor object
```
jacobweinstock, line 18: what is `mvcli`?

internal/sum/sum.go
```diff
@@ -0,0 +1,279 @@
+package sum
```
jacobweinstock, line 1: is `sum` an acronym?

internal/executor/executor.go
splaspood: It's external to the sum package with the expectation that there could/would be other similar utilities needing this in the future.

internal/sum/sum.go
splaspood: good catch, thank you.

internal/sum/sum.go
splaspood: Yes, Supermicro Update Manager https://www.supermicro.com/en/solutions/management-software/supermicro-update-manager

internal/executor/executor.go
jacobweinstock: what the the other utilities that will need and or implement this? if we dont have other uses we should hold off creating this.

This also is a big interface and its purpose and value add are not very clear. are we planning on replacing all calls to the `os/exec` with this? 

"Do not define interfaces before they are used: without a realistic example of usage, it is too difficult to see whether an interface is even necessary, let alone what methods it ought to contain." - https://go.dev/wiki/CodeReviewComments#interfaces

internal/sum/sum.go
jacobweinstock: thanks, please make a package level code comment so we dont lose this info once the PR is merged.

internal/sum/sum.go
```diff
@@ -0,0 +1,279 @@
...
+	e := ex.NewExecutor(sum.SumPath)
+	e.SetEnv([]string{"LC_ALL=C.UTF-8"})
+	// e.SetQuiet()
+	sum.Executor = e
+
+	return sum, nil
+}
+
+// Return a Fake sum executor for tests
+func NewFakeSum(r io.Reader) (*Sum, error) {
```
jacobweinstock, line 77: Any reason this can't live in a test package only?

internal/executor/executor.go
splaspood: I do not have any currently pending, so your argument about not implementing it before it's necessary is valid.   

That being said, this code was already implemented elsewhere for a very similar use case so it was adopted nearly as-is.    The presumption is that we'd need the currently unused functionality as we add more vendor utilities as we've seen in ironlib.

jacobweinstock: im not familiar with the ironlib code base. does ironlib already have the supermicro update manager code? should we just be importing from ironlib?

I'm also confused. you say "do not have any currently pending" and "as we add more vendor utilities". How can this be both?

internal/sum/sum.go
splaspood: Nope, and you're correct it's not used anywhere.   Bye bye.

internal/executor/executor.go
splaspood: Ironlib has many things, but not sum.   I was referring to the Executor/FakeExecutor stuff that's used for testing/execing in ironlib.

While I do not currently have any vendor utilities in the queue, I can envision a future with supporting other vendor's utilities where this code would be reused.

providers/sum/sum.go
splaspood: I'd like to handle this in a future PR.

joelrebel (approved):

internal/executor/executor.go
joelrebel: > should we just be importing from ironlib?
ironlib is mean't to execute vendor utilities in an OS running on the host.
bmclib is mean't to execute actions/calls remotely and so I think this fits in here

internal/executor/executor.go
joelrebel: when we're using a mock library, these would not be required

internal/executor/executor.go
splaspood: If we plan to continue using this code over the longer term vs something off the shelf as suggested by @DoctorVin , I'd argue that this should be extracted into an external package that both projects could import, but we haven't finalized those plans one way or the other at this point.

PR #394, opened 2024-08-20
Supermicro X13 inventory/firmware support FS-1671

splaspood:

Support for hw inventory and firmware management for Supermicro X13 boards.

Supermicro

X13

```
Supermicro X13 inventory/firmware support FS-1671
```

### DIFF
--- a/internal/redfishwrapper/firmware.go
+++ b/internal/redfishwrapper/firmware.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/bmc-toolbox/bmclib/v2/constants"
 	bmclibErrs "github.com/bmc-toolbox/bmclib/v2/errors"
+	redfish "github.com/stmcginnis/gofish/redfish"
 )
 
 type installMethod string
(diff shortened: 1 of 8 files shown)


providers/supermicro/supermicro.go
```diff
@@ -300,6 +320,11 @@ func (c *Client) bmcQueryor(ctx context.Context) (bmcQueryor, error) {
 			return nil, errors.Wrap(ErrModelUnknown, err.Error())
 		}
 
+		// ensure the device model matches the expected queryor
+		if !expected(deviceModel, bmc) {
```
joelrebel, line 324: include a `bmc.Close()` before the continue here to ensure we don't leave a successful connection open

providers/supermicro/supermicro.go
```diff
@@ -303,8 +334,8 @@ func (c *Client) bmcQueryor(ctx context.Context) (bmcQueryor, error) {
 	}
 
 	model := strings.ToLower(queryor.deviceModel())
-	if !strings.HasPrefix(model, "x12") && !strings.HasPrefix(model, "x11") {
-		return nil, errors.Wrap(ErrModelUnsupported, "expected one of X11* or X12*, got:"+model)
+	if !strings.HasPrefix(model, "x13") && !strings.HasPrefix(model, "x12") && !strings.HasPrefix(model, "x11") {
```
DoctorVin, line 337: With your new `expected` function I don't think you'll actually get here if it's not `x11`, `x12`, or `x13`. I think the queryor nil check @ 301 covers you.

internal/redfishwrapper/firmware.go
```diff
@@ -99,6 +100,14 @@ func (c *Client) FirmwareUpload(ctx context.Context, updateFile *os.File, params
 		)
 	}
 
+	// For X13 the full Task structure is returned in the body.   If we can Unmarshall then we can safely assume
+	// that redfishTask.ID contains the ID.
+	redfishTask := &redfish.Task{}
+	err = json.Unmarshal(response, redfishTask)
```
DoctorVin, line 106: Just a note, if response is valid JSON but isn't a `redfish.Task` I believe you'll get no error and zero value for the `Task`.  It might be worth a sanity check that the task id isn't an empty string here.

PR #395, opened 2024-09-19
port to gofish/redfish v0.19.0

DoctorVin:

As of v0.16.0 StartUpdateTarget is no longer exported from the redfish UpdateService. This breaks the build because we do a raw POST to this endpoint (to retrieve the Task object in the response body). The endpoint for initiating a firmware update via the UpdateService.StartUpdate REST endpoint is actually defined, however (section 6.141.4.4 cf. https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2024.1.pdf ). This change uses the defined endpoint instead of relying on a property returned by gofish's implementation.

SMC X12


coffeefreak101 (approved):

providers/asrockrack/helpers.go
```diff
@@ -578,7 +578,7 @@ func (a *ASRockRack) httpsLogin(ctx context.Context) error {
 func (a *ASRockRack) httpsLogout(ctx context.Context) error {
 	_, statusCode, err := a.queryHTTPS(ctx, "api/session", "DELETE", nil, nil, 0)
 	if err != nil {
-		return fmt.Errorf("Error logging out: " + err.Error())
+		return errors.Wrap(err, "logging out")
```
jacobweinstock, line 581: `github.com/pkg/errors` is actually archived. We should eventually remove its use entirely. You might think about using stdlib wrapping functionality for this. For example: `fmt.Errorf("logging out: %w", err)`

splaspood: I agree.   If it's not a big lift, I'd say let's do it here.   If it is, I'd like to see that in a follow-on PR.

splaspood (changes requested):
Other than that error wrapping discussion, lgtm.   We will have to test this extensively...  I'm concerned that we were pulling the start update endpoint from there for a reason other than it not simply being defined elsewhere... some weird vendor nonsense..   Joel likely knows.

providers/asrockrack/helpers.go
DoctorVin: I’m definitely on-board with this. I’ll revise tomorrow.

DoctorVin:
> Other than that error wrapping discussion, lgtm. We will have to test this extensively... I'm concerned that we were pulling the start update endpoint from there for a reason other than it not simply being defined elsewhere... some weird vendor nonsense.. Joel likely knows.

This particular method is only called from/used in the SMC X12 provider, so I am less concerned we’ll break unknown providers. That said, I agree that we need to be vigilant here. I am suspicious that redfish is going to be nice with older vintages of BMC software.

joelrebel:
> I am suspicious that redfish is going to be nice with older vintages of BMC software.

That would make our lives easier, but vendors often implement their own interpretation of redfish.
The action endpoints are ideally retrieved instead of hardcoded, given that its just SMC X12 using the method we let this through.

I'm more concerned about the extensive changes that have been introduced since we last updated Gofish - https://github.com/stmcginnis/gofish/compare/v0.15.0...v0.19.0

I suggest holding off the merge until we've tested this with some of our tools (PBnJ, Alloy, Flasher) with this rev on some hardware. One of the main concerns I had was the concurrent processing added here, and how that affects the inventory collection https://github.com/stmcginnis/gofish/commit/a9703efaf43745cd4185c75d5dc7a66817e22f53

That hardware test pipeline would have been nice to have here.

providers/asrockrack/helpers.go
DoctorVin: I did this for the code I changed, but left the existing usage.

splaspood (approved):

joelrebel (approved):

PR #398, opened 2024-10-08
Update virtual media mounting:

jacobweinstock:

At least one BMC, Supermicro X11SDV-4C-TLN2F, did not support setting `inserted` and/or `writeProtected` in the payload properties in redfish calls to do a virtual media mount. This falls back to not using them if the initial call with them in the properties fails.

This was test and worked successfully on a Supermicro (X11SDV-4C-TLN2F), HP ILO5, and Dell iDRAC9.

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/internal/redfishwrapper/virtual_media.go
+++ b/internal/redfishwrapper/virtual_media.go
@@ -48,22 +48,20 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
 					return false, err
 				}
 			}
-			if mediaURL != "" {
-				if slices.Contains(vm.MediaTypes, mediaKind) {
-					if err := vm.InsertMedia(mediaURL, true, true); err != nil {
-						// Some BMC's (Supermicro SYS-E300-D9, for example) don't support the "inserted" and "writeProtected" properties,
-						// so we try to insert the media without them if the first attempt fails.
-						if err := vm.InsertMediaConfig(rf.VirtualMediaConfig{Image: mediaURL}); err != nil {
-							return false, err
-						}
-					}
-					return true, nil
-				}
-
+			if mediaURL == "" {
+				// Only ejecting the media was requested.
+				return true, nil
+			}
+			if !slices.Contains(vm.MediaTypes, mediaKind) {
 				return false, fmt.Errorf("media kind %s not supported by BMC, supported media kinds %q", kind, vm.MediaTypes)
 			}
-
-			// Only ejecting the media was requested.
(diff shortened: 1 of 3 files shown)


internal/redfishwrapper/virtual_media.go
```diff
@@ -29,50 +33,43 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
...
+					return false, err
+				}
 			}
-
-			for _, media := range virtualMedia {
-				for _, t := range media.MediaTypes {
-					if t == mediaKind {
-						err = media.InsertMedia(mediaURL, true, true)
-						if err != nil {
+			if mediaURL != "" {
```
joelrebel, line 51: could this return early by inverting the check? that would make it easier to read as well

internal/redfishwrapper/virtual_media.go
```diff
@@ -29,50 +33,43 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
...
+				if slices.Contains(vm.MediaTypes, mediaKind) {
```
joelrebel, line 52: similar here, return early for readability

internal/redfishwrapper/virtual_media.go
```diff
@@ -29,50 +33,43 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
...
+					if err := vm.InsertMedia(mediaURL, true, true); err != nil {
+						// Some BMC's (Supermicro SYS-E300-D9, for example) don't support the "inserted" and "writeProtected" properties,
```
joelrebel, line 54: Can this comment be updated to to mainboard model number instead - X11SDV-4C-TLN2F,
The product name - SYS-* tends to change based on where the hardware is purchased and bmclib internally uses the mainboard model.

joelrebel (commented):
Thanks for adding SMC support and testing this out, would you mind adding in an example snipped for reference into https://github.com/bmc-toolbox/bmclib/tree/main/examples

internal/redfishwrapper/virtual_media.go
jacobweinstock: ah yes, sorry about that. I will update it. thanks for looking up the mainboard model number!

jacobweinstock:
@joelrebel , ready for another review. please and thank you!

joelrebel (approved):
Thanks, thanks looks good to me

PR #399, opened 2024-10-17
expand response payload processing

DoctorVin:
In contrast to other server-vendors, SMC does not return the task id in the Location header of the response to a firmware upload. In BMC version 1.05.03 (Redfish version 1.14.0) the payload format changes from a TaskAccepted message to a Redfish task, which breaks task id detection. This change adds an attempt to deserialize the task structure before falling back to the earlier TaskAccepted message type.

This also corrects the startUpdateURI.

### DIFF
--- a/internal/redfishwrapper/firmware.go
+++ b/internal/redfishwrapper/firmware.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


ofaurax (approved):

jakeschuurmans (commented):
fix linter and we good

PR #400, opened 2024-10-21
Allow specifying the Redfish system name:

jacobweinstock:
For Redfish implementations that manage multiple systems being able to specify the system name is required to be able to interact with the desired machine. This is the case with sushy-tools, which is a redfish emulator for libvirt and others. Also, enabling redfish basic auth was not plumbed through and is now.

### Checklist
- [ ] Tests added

```
```


internal/redfishwrapper/system.go
```diff
@@ -35,7 +35,12 @@ func (c *Client) Systems() ([]*redfish.ComputerSystem, error) {
 		return nil, err
 	}
 
-	return c.client.Service.Systems()
+	s, err := c.client.Service.Systems()
+	if err != nil {
+		return nil, err
+	}
+
+	return c.matchingSystem(s), nil
```
joelrebel, line 43: We'd want to keep the default behavior here - to return all Systems - when c.systemName was not specified

```suggestion
        if c.systemName != "" {
	    return c.matchingSystem(s), nil
	}
	
	return s, nil
```

providers/redfish/redfish.go
```diff
@@ -120,6 +127,8 @@ func New(host, user, pass string, log logr.Logger, opts ...Option) *Conn {
 		redfishwrapper.WithHTTPClient(defaultConfig.HttpClient),
 		redfishwrapper.WithVersionsNotCompatible(defaultConfig.VersionsNotCompatible),
 		redfishwrapper.WithEtagMatchDisabled(defaultConfig.DisableEtagMatch),
+		redfishwrapper.WithBasicAuthEnabled(defaultConfig.UseBasicAuth),
```
joelrebel, line 130: thanks for fixing this

joelrebel (approved):

PR #402, opened 2024-10-28
providers/supermicro/supermicro.go: Initialize sum 'client' during serviceclient init

splaspood:

A regression crept in recently where the 'sum' client was not being assigned to the 'serviceclient' when the supermicro provider was instantiated.     This change fixes that problem.

### DIFF
--- a/providers/supermicro/supermicro.go
+++ b/providers/supermicro/supermicro.go
@@ -136,13 +136,17 @@ func NewClient(host, user, pass string, log logr.Logger, opts ...Option) *Client
 		opt(defaultConfig)
 	}
 
-	serviceClient := newBmcServiceClient(
+	serviceClient, err := newBmcServiceClient(
 		host,
 		defaultConfig.Port,
 		user,
 		pass,
 		httpclient.Build(defaultConfig.httpClientSetupFuncs...),
 	)
+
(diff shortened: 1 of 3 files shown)


jakeschuurmans (approved):

providers/supermicro/supermicro.go
```diff
@@ -136,14 +136,18 @@ func NewClient(host, user, pass string, log logr.Logger, opts ...Option) *Client
...
+	serviceClient, err := newBmcServiceClient(
 		host,
 		defaultConfig.Port,
 		user,
 		pass,
 		httpclient.Build(defaultConfig.httpClientSetupFuncs...),
 	)
 
+	if err != nil {
+		return nil
```
joelrebel, line 148: Is this intentionally returning nil here?

splaspood: Yeah, ultimately in this case I was unsure what to do with the error...   I suppose an argument could be made that this is a marginal improvement to error handling in this case compared to previous, but that's likely a bit generous.

joelrebel: ack, I'd suggest leaving a comment to indicate its intentional

jakeschuurmans (approved):

PR #403, opened 2024-11-08
Fix SetBiosFromFile

jakeschuurmans:
Fix example

Fix SetBiosFromFile
Fix the SetBiosFromFile example


joelrebel (approved):
ideally this interface could have accepted an io.Reader

PR #408, opened 2024-12-14
Nil pointer safety

jacobweinstock:
Fix bug introduced by #402. This PR handles errors properly and adds some nil checking in the `bmc` package. By ignoring an error and returning nil objects it causes`panic: runtime error: invalid memory address or nil pointer dereference` for clients.

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/providers/supermicro/supermicro.go
+++ b/providers/supermicro/supermicro.go
@@ -136,20 +136,13 @@ func NewClient(host, user, pass string, log logr.Logger, opts ...Option) *Client
 		opt(defaultConfig)
 	}
 
-	serviceClient, err := newBmcServiceClient(
+	serviceClient := newBmcServiceClient(
 		host,
 		defaultConfig.Port,
 		user,
 		pass,
 		httpclient.Build(defaultConfig.httpClientSetupFuncs...),
 	)
-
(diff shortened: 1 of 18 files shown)


providers/supermicro/supermicro.go
```diff
@@ -450,20 +443,27 @@ type serviceClient struct {
...
 
-func newBmcServiceClient(host, port, user, pass string, client *http.Client) (*serviceClient, error) {
+func newBmcServiceClient(host, port, user, pass string, client *http.Client) *serviceClient {
 	sc := &serviceClient{host: host, port: port, user: user, pass: pass, client: client}
 
 	if !strings.HasPrefix(host, "https://") && !strings.HasPrefix(host, "http://") {
 		sc.host = "https://" + host
 	}
 
+	// sum is only for firmware related operations. Failing the client entirely because of a sum error
```
joelrebel, line 453: The sum code can be removed entirely since bmclib was 'hard' forked (for the environment where sum was required to run remotely). I'll get round to cleaning this up sometime soon.

joelrebel (approved):
Thanks for fixing this up

PR #412, opened 2025-01-17
Improve virtual media ejecting and inserting:

jacobweinstock:

As there can be multiple virtual media for a single manager, i.e. CD, DVD, USBStick, etc, we need to check for a matching supported type before ejecting or injecting. Also, added checking that the virtual media supports injecting before trying.

### Checklist
- [ ] Tests added

```
```


joelrebel (approved):
I suggest implementing tests to avoid regressions for the various cases being handled here

PR #414, opened 2025-03-11
Tooling updates + add Proxy support to httpclient.DefaultTransport

mmlb:

- Some spring cleaning of non-source files
- Addition of Proxy config to the httpclient.DefaultTransport
- Add example binaries to .gitignore

Spring cleaning changes are annoyingly large but mostly mechanical, either via `gofumpt` or `golangci-lint run --fix`, there were some manual fixes I made to avoid disabling linters that complained about a handful of things.

I also added Proxy config to the DefaultTransport just like the DefaultTransport in the http package does. This setup will make it so http requests will go via a proxy when HTTPS_PROXY env var exists. This way we can debug redfish over https using mitmproxy or similar setup.

I added the binaries to .gitignore because they should have been there from the beginning.

```
https connections now go through a proxy if the HTTPS_PROXY environment variable is set
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -208,7 +208,6 @@ func (c *Client) registerGofishProvider() {
 
 // register Intel AMT provider
 func (c *Client) registerIntelAMTProvider() {
-
 	iamtOpts := []intelamt.Option{
 		intelamt.WithLogger(c.Logger),
 		intelamt.WithHostScheme(c.providerConfig.intelamt.HostScheme),
@@ -221,7 +220,7 @@ func (c *Client) registerIntelAMTProvider() {
 // register Dell gofish provider
 func (c *Client) registerDellProvider() {
 	dellGofishHttpClient := *c.httpClient
-	//dellGofishHttpClient.Transport = c.httpClient.Transport.(*http.Transport).Clone()
+	// dellGofishHttpClient.Transport = c.httpClient.Transport.(*http.Transport).Clone()
 	dellGofishOpts := []dell.Option{
 		dell.WithHttpClient(&dellGofishHttpClient),
 		dell.WithVersionsNotCompatible(c.providerConfig.dell.VersionsNotCompatible),
@@ -373,7 +372,6 @@ func (c *Client) Open(ctx context.Context) error {
 
 // Close pass through to library function
 func (c *Client) Close(ctx context.Context) (err error) {
-
 	ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "Close")
 	defer span.End()
 
@@ -613,7 +611,6 @@ func (c *Client) FirmwareInstallStatus(ctx context.Context, installVersion, comp
 	metadata.RegisterSpanAttributes(c.Auth.Host, span)
 
 	return status, err
-
 }
 
 // PostCodeGetter pass through library function to return the BIOS/UEFI POST code
--- a/client.go
+++ b/client.go
@@ -501,7 +501,7 @@ func (c *Client) SetBootDevice(ctx context.Context, bootDevice string, setPersis
 // server. Specifically, the method ejects any currently attached virtual media, and then if
 // mediaURL isn't empty, attaches a virtual media device of type kind whose contents are
 // streamed from the indicated URL.
-func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL string) (ok bool, err error) {
+func (c *Client) SetVirtualMedia(ctx context.Context, kind, mediaURL string) (ok bool, err error) {
 	ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "SetVirtualMedia")
 	defer span.End()
 
@@ -588,7 +588,7 @@ func (c *Client) ResetBiosConfiguration(ctx context.Context) (err error) {
 }
 
 // FirmwareInstall pass through library function to upload firmware and install firmware
-func (c *Client) FirmwareInstall(ctx context.Context, component string, operationApplyTime string, forceInstall bool, reader io.Reader) (taskID string, err error) {
+func (c *Client) FirmwareInstall(ctx context.Context, component, operationApplyTime string, forceInstall bool, reader io.Reader) (taskID string, err error) {
 	ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "FirmwareInstall")
 	defer span.End()
 
@@ -599,7 +599,7 @@ func (c *Client) FirmwareInstall(ctx context.Context, component string, operatio
 	return taskID, err
 }
 
-// Note: this interface is to be deprecated in favour of a more generic FirmwareTaskStatus.
+// Note: this interface is to be deprecated in favor of a more generic FirmwareTaskStatus.
 //
 // FirmwareInstallStatus pass through library function to check firmware install status
 func (c *Client) FirmwareInstallStatus(ctx context.Context, installVersion, component, taskID string) (status string, err error) {
--- a/client.go
+++ b/client.go
@@ -148,12 +148,20 @@ func (c *Client) defaultTimeout(ctx context.Context) time.Duration {
 	return time.Until(deadline) / time.Duration(l)
 }
 
+func cloneHTTPClient(old *http.Client) *http.Client {
+	client := *old
+	transport, ok := client.Transport.(*http.Transport)
+	if !ok {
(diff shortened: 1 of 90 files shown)


internal/redfishwrapper/task.go
```diff
@@ -76,13 +74,13 @@ func (c *Client) ConvertTaskState(state string) constants.TaskState {
 	switch strings.ToLower(state) {
 	case "starting", "downloading", "downloaded", "scheduling":
 		return constants.Initializing
-	case "running", "stopping", "cancelling":
+	case "running", "stopping", "canceling":
```
jacobweinstock, line 77: I don't think these should change as they are States that correspond to the gofish library. https://github.com/stmcginnis/gofish/blob/4b51c15f0a3df649821f9fc395a82483ff60bb2a/redfish/task.go#L16-L55 

It is spelled `Cancelling` in that library. We should probably be using the const from that package anyway.

client.go
```diff
@@ -148,12 +148,20 @@ func (c *Client) defaultTimeout(ctx context.Context) time.Duration {
 	return time.Until(deadline) / time.Duration(l)
 }
 
+func cloneHTTPClient(old *http.Client) *http.Client {
+	client := *old
+	transport, ok := client.Transport.(*http.Transport)
+	if !ok {
+		panic("old client's transport is not expected type")
```
jacobweinstock, line 155: hmm...I'm concerned about this panic. I know we weren't even checking the type assert before, and it could have implicitly caused a panic already, but could we introduce a more graceful way to handle the bad type assert? Maybe a return error or something similar?

jacobweinstock (changes requested):

client.go
mmlb: I don't think its worth it, changing to an error return is going to have wide effect when done for `registerASRRProvider` and probably others (bubbling the error up). Its never really going to be a problem either since we aren't really using different implementations. If it ever changes then it'll be caught early at startup instead of later at runtime.

Not a whole lot of benefit imo.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bmc-toolbox/bmclib/v2
 
-go 1.21
+go 1.23
```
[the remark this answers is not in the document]
mmlb: Right, but no one should be running go <1.23 anyway. I kind of see it as a chicken/egg situation. When 1.25 comes out I'd like to bump this to 1.24 and then make use of `go tool` support to run various tools (like golangci-lint) via native go pinning.

internal/redfishwrapper/task.go
mmlb: uggh good catch will back out. We are using the consts from that package already. The problem is we are converting that package's types <-> string when it should be done by that package itself (via stringer imo).

providers/asrockrack/helpers.go
```diff
@@ -602,7 +600,7 @@ func (a *ASRockRack) queryHTTPS(ctx context.Context, endpoint, method string, pa
 	}
 
 	// add headers
-	req.Header.Add("X-CSRFTOKEN", a.loginSession.CSRFToken)
+	req.Header.Add("X-Csrftoken", a.loginSession.CSRFToken)
```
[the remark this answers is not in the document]
mmlb: this was an autofix from one of the linters, can do the magic comment to have it ignore this but I'd like some clarity before going that route.

I'm not seeing anything in the code base about old firmware expecting specific case of headers, are you thinking about supermicro with form-data fields https://github.com/bmc-toolbox/bmclib/blob/main/providers/supermicro/floppy.go#L91-L96?

go.mod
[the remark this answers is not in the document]
jacobweinstock: Sorry, that's not the [policy](https://github.com/bmc-toolbox/bmclib?tab=readme-ov-file#go-version-in-gomod). As a library, we shouldn't dictate the version clients use. Per the policy, `go tool` support isn't a reason to update this. There are many examples of older Go versions in large active Go module repos in the wild. https://github.com/spf13/cobra, for example.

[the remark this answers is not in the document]
mmlb: oh til! well will back this out then

client.go
joelrebel: This does seem like an exceptional case, can the panic message include a clearer message, along the lines of "expected a http client of the http.Transport type, got $blah instead"?

joelrebel (changes requested):
Thanks for updates and fixes @mmlb , for the reviewers sake, going ahead could we keep the PRs scope limited

providers/asrockrack/helpers.go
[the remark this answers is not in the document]
joelrebel: I'd recommend the headers are left as is, unless a change is required in practice

mmlb:
> Thanks for updates and fixes @mmlb , for the reviewers sake, going ahead could we keep the PRs scope limited

I'm going to drop new features from this PR and just do the format/lint fixes, aka the cleaning part

go.mod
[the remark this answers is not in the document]
mmlb: there's also the `copyloopvar` change that came in from 1.22 that affects this repo (albeit just one line). I'll back this out and have a separate PR for 1.22 + enabling those linters.

providers/asrockrack/helpers.go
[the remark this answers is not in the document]
mmlb: will do

go.mod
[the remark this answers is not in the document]
mmlb: Alright dropped back down to 1.21 and undid changes

providers/asrockrack/helpers.go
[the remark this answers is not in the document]
mmlb: should be back to original format

internal/redfishwrapper/task.go
mmlb: should be back to original (plus `//nolint:` directive)

client.go
mmlb: I've updated the panic message

mmlb:
> Thanks for updates and fixes @mmlb , for the reviewers sake, going ahead could we keep the PRs scope limited

I've trimmed things down and also split the commits a little bit so that they are easier to review. I split out what was previously "automatic fixes + manual fixes" into multiple ones. The first iteration was all together so that each commit would still pass linting so it was quite large. Now a couple of intermediate commits won't pass but makes for smaller diffs.

joelrebel (approved):
This looks fine to me

providers/supermicro/firmware_bios_test.go
```diff
@@ -93,7 +94,7 @@ func Test_setComponentUpdateMisc(t *testing.T) {
 					return
 				}
 
-				assert.Nil(t, err)
+				assert.NoError(t, err)
```
jacobweinstock, line 97: The use of `require.NoError` and `assert.NoError` seems arbitrary. When to use which is not clear. Is this something that `golangci-lint` catches? If not, I recommend either creating a doc for when one should be used over the other, or just stick to one.

internal/redfishwrapper/task.go
```diff
@@ -97,6 +95,8 @@ func (c *Client) TaskStateActive(state constants.TaskState) (bool, error) {
 		return true, nil
 	case constants.Complete, constants.Failed:
 		return false, nil
+	case constants.PowerCycleHost, constants.Unknown:
+		fallthrough
```
jacobweinstock, line 99: For what is this optimizing? Why is this needed if it's just going to fallthrough to the default? Won't the default case be called regardless of whether this case exists or not?

jacobweinstock (commented):
> Thanks for updates and fixes @mmlb , for the reviewers sake, going ahead could we keep the PRs scope limited

I don't think multiple commits was what @joelrebel was driving at here. Limiting scope would be, for example, creating a PR just for the linting changes, one for the small pedantic differences you prefer, and another for just the proxy support.

internal/redfishwrapper/task.go
jacobweinstock: Oh, is this to satisfy a linting complaint?

joelrebel: `constants.PowerCycleHost` and `constants.Unknown` should not be part of the switch case since they are not Redfish task states,
could we have this case removed?

http://redfish.dmtf.org/schemas/DSP2046_2019.4.html#redfish.dmtf.org/schemas/v1/Task.json|details|TaskState

joelrebel:
Closing this PR as its not being worked on anymore

PR #416, opened 2025-03-26
Removed power state limitations from RPC provider

MrFlynn:
Since the RPC provider can be any arbitrary backend, the specific backend implementation should be responsible for handling any unsupported power state (by returning an error), not this library itself.

### Checklist
- [ ] ~~Tests added~~ All existing tests pass.
- [ ] ~~Similar commits squashed~~

```
Removed power state limitations for RPC provider. All possible power states, not just "on", "off", or "cycle" are now supported by this provider.
```

### DIFF
--- a/providers/rpc/rpc.go
+++ b/providers/rpc/rpc.go
@@ -3,7 +3,6 @@ package rpc
 import (
 	"bytes"
 	"context"
-	"errors"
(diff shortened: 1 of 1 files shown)


jacobweinstock (approved):
Thanks, @MrFlynn

Issue #421, opened 2025-10-01
matchingSystem returns all systems if there is no system matching systemName


willemm:
At the moment, the function matchingSystem will return a list of all systems if none of the system matches.

https://github.com/bmc-toolbox/bmclib/blob/6fa3269c7651473272b1e2b0e9d8b9c60532f6a0/internal/redfishwrapper/system.go#L81

This means that if you make a typo in your system name, or if it isn't present at some point, then this library will happily power down *all* of the systems it has access to.

PR #422, opened 2025-10-01
Don't use all systems when no match is found:

jacobweinstock:
Don't use all systems when no match is found. This will cause a not found system to return all systems. This will not allow users to interact with all systems by not setting a system name. This changes the behavior of the API but is/was the intended behavior.

fixes: #421 

### Checklist
- [x] Tests added

```
```

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,5 +1,6 @@
 queue_rules:
   - name: default
+    batch_size: 1
     queue_conditions:
       - base=main
       - "#approved-reviews-by>=1"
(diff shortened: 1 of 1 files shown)


rpardini:
Thanks. I've hit the case where systemName didn't match and got half a DC powered down. This of course only happened when talking to a multi-System Redfish (eg, OpenManage Enterprise) and not a single/individual iDRAC.

joelrebel (approved):
Thanks for fixing this up

PR #438, opened 2026-05-27
Upgrade gofish to v0.22.0 and add workaround for Lenovo VirtualMedia …

nuxster:

This PR updates the `gofish` dependency to v0.22.0 and adapts `bmclib`'s Virtual Media handling to leverage new fallback mechanisms.

### Key changes:
- **Virtual Media PATCH Fallback:** `bmclib` now supports `InsertMedia` and `EjectMedia` operations via `PATCH` requests for BMCs that do not expose standard Redfish Actions for Virtual Media (e.g., Lenovo XCC). This is achieved by relying on `gofish v0.22.0`'s ability to fallback to PATCHing the VirtualMedia resource directly.
- **Safer Field Access:** Uses `gofish.Deref` for safer access to optional fields like `Inserted`, preventing potential panics if the field is not returned by the BMC.
- **Updated Tests:** New unit tests (`TestSetVirtualMedia_InsertPatchFallbackWithoutAction`, `TestSetVirtualMedia_EjectPatchFallbackWithoutAction`) have been added to specifically cover the new `PATCH` fallback behavior.
- **Documentation and Examples:** `internal/redfishwrapper/doc.go` includes a new section on Virtual Media, detailing the `PATCH` fallback and the `-eject` flag. The `examples/virtualmedia/doc.go` has been updated to use the `-eject` flag and explicitly select the `gofish` provider.

### Checklist
- [x] Tests added
- [ ] Similar commits squashed (Assuming this PR is a single logical change, no complex squashing needed unless requested)

Lenovo (specifically XCC mentioned in context), and potentially other BMCs that do not expose standard Redfish Virtual Media Actions.

Not specified, but context points to Lenovo XCC.

BMC firmware versions compatible with `gofish v0.22.0` that lack standard Virtual Media Actions.

- `gofish` v0.22.0

Updated `gofish` dependency to v0.22.0. `bmclib` now supports inserting and ejecting virtual media via `PATCH` requests for BMCs that do not expose standard Redfish Virtual Media Actions (e.g., Lenovo XCC). This fallback mechanism ensures broader compatibility. Added new unit tests for this functionality and updated documentation and examples accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,46 @@ client := bmclib.NewClient("192.168.1.1", "admin", "hunter2", opt...)
 cl.Registry.Drivers = cl.Registry.FilterForCompatible(ctx)
 ```
 
+## Virtual Media
+
+bmclib supports attaching and ejecting virtual media through the Redfish `gofish` provider.
+
+```go
+cl := bmclib.NewClient("192.168.1.1", "admin", "hunter2")
+
+if err := cl.Open(ctx); err != nil {
+  return(err)
+}
+defer cl.Close(ctx)
+
+// Attach an ISO image as CD/DVD virtual media.
+ok, err := cl.SetVirtualMedia(ctx, "CD", "https://example.com/installer.iso")
+if err != nil {
+  return(err)
+}
+if !ok {
+  return(fmt.Errorf("failed to set virtual media"))
+}
+
+// To eject currently attached virtual media, pass an empty media URL:
+ok, err := cl.SetVirtualMedia(ctx, "CD", "")
+if err != nil {
+  return(err)
+}
+if !ok {
+  return(fmt.Errorf("failed to eject virtual media"))
+}
+```
+
+Supported media kinds are:
+* CD
+* DVD
+* Floppy
+* USBStick
+
+The Redfish provider first looks for `VirtualMedia` under the `Manager` resource (/redfish/v1/Managers/{ManagerId}/VirtualMedia) and falls back to the `System` resource (/redfish/v1/Systems/{SystemId}/VirtualMedia) for implementations that expose virtual media there.
+When the BMC does not expose standard `VirtualMedia.InsertMedia` or `VirtualMedia.EjectMedia` actions, bmclib relies on gofish to fall back to PATCHing the `VirtualMedia` resource directly. This allows virtual media support on implementations such as Lenovo XCC that manage virtual media through `PATCH` instead of Redfish Actions.
+
 ## Timeouts
 
 bmclib can be configured to apply timeouts to BMC interactions. The following options are available.
--- a/README.md
+++ b/README.md
@@ -121,46 +121,6 @@ client := bmclib.NewClient("192.168.1.1", "admin", "hunter2", opt...)
 cl.Registry.Drivers = cl.Registry.FilterForCompatible(ctx)
 ```
 
-## Virtual Media
-
-bmclib supports attaching and ejecting virtual media through the Redfish `gofish` provider.
-
-```go
-cl := bmclib.NewClient("192.168.1.1", "admin", "hunter2")
-
-if err := cl.Open(ctx); err != nil {
-  return(err)
-}
-defer cl.Close(ctx)
-
-// Attach an ISO image as CD/DVD virtual media.
-ok, err := cl.SetVirtualMedia(ctx, "CD", "https://example.com/installer.iso")
-if err != nil {
-  return(err)
-}
-if !ok {
-  return(fmt.Errorf("failed to set virtual media"))
-}
-
-// To eject currently attached virtual media, pass an empty media URL:
-ok, err := cl.SetVirtualMedia(ctx, "CD", "")
-if err != nil {
-  return(err)
-}
-if !ok {
-  return(fmt.Errorf("failed to eject virtual media"))
-}
-```
-
-Supported media kinds are:
-* CD
-* DVD
-* Floppy
-* USBStick
-
-The Redfish provider first looks for `VirtualMedia` under the `Manager` resource (/redfish/v1/Managers/{ManagerId}/VirtualMedia) and falls back to the `System` resource (/redfish/v1/Systems/{SystemId}/VirtualMedia) for implementations that expose virtual media there.
-When the BMC does not expose standard `VirtualMedia.InsertMedia` or `VirtualMedia.EjectMedia` actions, bmclib relies on gofish to fall back to PATCHing the `VirtualMedia` resource directly. This allows virtual media support on implementations such as Lenovo XCC that manage virtual media through `PATCH` instead of Redfish Actions.
-
 ## Timeouts
 
 bmclib can be configured to apply timeouts to BMC interactions. The following options are available.
--- a/examples/virtualmedia/main.go
+++ b/examples/virtualmedia/main.go
@@ -20,13 +20,19 @@ func main() {
 	user := flag.String("user", "", "BMC username, required")
 	pass := flag.String("password", "", "BMC password, required")
 	host := flag.String("host", "", "BMC hostname or IP address, required")
-	isoURL := flag.String("iso", "", "The HTTP URL to the ISO/image to be mounted")
+	isoURL := flag.String("iso", "", "The HTTP URL to the ISO/image to be mounted; not valid with -eject")
 	mediaKind := flag.String("media-kind", "CD", "Virtual media kind: CD, DVD, Floppy, USBStick")
 	eject := flag.Bool("eject", false, "Eject/unmount virtual media instead of mounting")
 	flag.Parse()
 
 	if *user == "" || *pass == "" || *host == "" {
 		fmt.Fprintln(os.Stderr, "user, password, and host are required")
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
+
+	if *eject && *isoURL != "" {
+		fmt.Fprintln(os.Stderr, "iso must not be supplied when -eject is set")
 		flag.PrintDefaults()
 		os.Exit(1)
 	}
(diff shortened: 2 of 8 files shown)


internal/redfishwrapper/virtual_media.go
```diff
@@ -89,34 +89,41 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
...
+			},
 		}
 
-		insertMedia := schemas.VirtualMediaInsertMediaParameters{
-			Image:          mediaURL,
-			Inserted:       gofish.ToRef(true),
-			WriteProtected: gofish.ToRef(true),
-		}
-		if _, err := vm.InsertMedia(&insertMedia); err != nil {
-			// Some BMCs (e.g., Supermicro X11SDV-4C-TLN2F) don't support the
```
joelrebel, line 110: Would prefer we keep these comments in the new changes on line 110

README.md
```diff
@@ -121,6 +121,46 @@ client := bmclib.NewClient("192.168.1.1", "admin", "hunter2", opt...)
 cl.Registry.Drivers = cl.Registry.FilterForCompatible(ctx)
 ```
 
+## Virtual Media
```
joelrebel: This section can be removed from the main README.md which is fairly large as it is, it would prove more helpful in the doc.go

internal/redfishwrapper/virtual_media.go
```diff
@@ -89,34 +89,41 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
...
 				slotErrors = append(slotErrors, fmt.Errorf("%s: eject before insert: %w", vm.ODataID, err))
 				continue
 			}
 		}
 
-		if !vm.SupportsMediaInsert {
-			slotErrors = append(slotErrors, fmt.Errorf("%s: does not support insert", vm.ODataID))
-			continue
+		insertMediaAttempts := []schemas.VirtualMediaInsertMediaParameters{
+			{
```
joelrebel, line 105: would it be worth having test cases to cover for these attempts?

joelrebel (changes requested):
Thanks for the efforts and contribution here, left a few suggestions.

internal/redfishwrapper/virtual_media.go
nuxster: I think yes,  the new tests 'TestSetVirtualMedia_InsertPatchFallbackRequiresInsertedWithoutWriteProtected' and 'TestSetVirtualMedia_EjectPatchFallbackWithoutAction' specifically cover these attempts by simulating scenarios where the first attempt fails and subsequent attempts (including the one with Inserted but without WriteProtected) succeed.

internal/redfishwrapper/virtual_media.go
nuxster: Fixed

README.md
nuxster: Fixed

internal/redfishwrapper/virtual_media.go
nuxster: In the future, I plan to transfer this experience to the development of a provider for Lenovo. These tests may come in handy later.

nuxster: I can delete the tests if you really think they are superfluous.

joelrebel: Those look fine to me, thanks

joelrebel:
This looks good to me
@jacobweinstock if you can have a quick look here

joelrebel (dismissed):

internal/redfishwrapper/virtual_media.go
```diff
@@ -89,34 +89,43 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
 		}
 
 		// Ejecting the media before inserting new media makes the success rate of inserting the new media higher.
-		if *vm.Inserted && vm.SupportsMediaEject {
+		if gofish.Deref(vm.Inserted) {
```
jacobweinstock, line 96: what's the motivation for removing, "vm.SupportsMediaEject"?

internal/redfishwrapper/virtual_media.go
```diff
@@ -89,34 +89,43 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
...
 			if _, err := vm.EjectMedia(); err != nil {
 				slotErrors = append(slotErrors, fmt.Errorf("%s: eject before insert: %w", vm.ODataID, err))
 				continue
 			}
 		}
 
-		if !vm.SupportsMediaInsert {
```
jacobweinstock, line 99: whats the motivation for removing this check?

internal/redfishwrapper/virtual_media.go
nuxster: Taking into account the innovations in the latest version of gofish, the Inserted field may be optional. If BMC does not return this field in the response, "vm.Inserted" will be nil. Direct dereference of "*vm.Inserted" in this case will result in "runtime error: invalid memory address or nil pointer dereference".
Also, the old logic required that BMC explicitly support "SupportsMediaEject" (through the presence of the corresponding Action in the Redfish API). However, as it turned out with Lenovo XCC and other BMCs, they may not provide the standard "Action#VirtualMedia.EjectMedia", but at the same time they allow you to execute an eject through a PATCH request to the VirtualMedia resource itself. The old check "&& vm.SupportsMediaEject" blocked this fallback mechanism.

internal/redfishwrapper/virtual_media.go
nuxster: The old logic assumed that if BMC does not provide an explicit Media Insertion Action (#VirtualMedia.InsertMedia), then the insertion operation is not supported. It was too strict a restriction.
Since gofish v0.22.0, if the standard Action "#VirtualMedia.InsertMedia" is missing, the library now tries to perform the "InsertMedia" operation using a PATCH request to the "VirtualMedia" resource itself. This allows you to work with BMCs (for example, Lenovo XCC) that do not implement standard Redfish Actions for this operation, but support PATCH management.

internal/redfishwrapper/virtual_media.go
jacobweinstock: Thanks @nuxster. This is fantastic context. Mind adding a very targeted version of this as code comments? So much of the learning and experience we gain when actually making calls against different vendor's is often lost. Code comments like will be incredibly valuable to future us.

internal/redfishwrapper/virtual_media.go
jacobweinstock: Thanks @nuxster. This is fantastic context. Mind adding a very targeted version of this as code comments? So much of the learning and experience we gain when actually making calls against different vendor's is often lost. Code comments like will be incredibly valuable to future us.

internal/redfishwrapper/virtual_media.go
nuxster: Of course, I will now add the relevant comments to the code.

nuxster: Done

internal/redfishwrapper/virtual_media.go
nuxster: Done

internal/redfishwrapper/virtual_media.go
```diff
@@ -89,34 +93,49 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind string, mediaURL stri
 		}
 
 		// Ejecting the media before inserting new media makes the success rate of inserting the new media higher.
-		if *vm.Inserted && vm.SupportsMediaEject {
+		if gofish.Deref(vm.Inserted) {
 			if _, err := vm.EjectMedia(); err != nil {
 				slotErrors = append(slotErrors, fmt.Errorf("%s: eject before insert: %w", vm.ODataID, err))
 				continue
 			}
 		}
```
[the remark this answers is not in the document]
nuxster: Fixed. If pre-insert eject fails, we now keep the error for diagnostics but still attempt InsertMedia for the same slot.

examples/virtualmedia/main.go
```diff
@@ -29,16 +31,30 @@ func main() {
...
 	}
 	defer cl.Close(ctx)
 
-	ok, err := cl.SetVirtualMedia(ctx, "CD", *isoURL)
+	mediaURL := *isoURL
+	operation := "mount"
+	if *eject {
+		mediaURL = ""
+		operation = "eject"
+	}
```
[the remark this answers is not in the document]
nuxster: Fixed. The example now treats `-eject` together with a non-empty `-iso` as invalid input and exits with an error instead of silently ignoring the ISO URL.

internal/redfishwrapper/virtual_media_test.go
```diff
@@ -219,6 +221,254 @@ func TestSetVirtualMedia_SlotFallback(t *testing.T) {
...
+
+	err = client.Open(ctx)
+	require.NoError(t, err)
+
+	defer client.Close(ctx)
+
+	ok, err := client.SetVirtualMedia(ctx, "CD", "http://example.com/boot.iso")
+	assert.NoError(t, err)
+	assert.True(t, ok)
+	assert.True(t, patchCalled, "expected InsertMedia fallback to PATCH the VirtualMedia resource")
```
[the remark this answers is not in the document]
nuxster: Fixed. I replaced the boolean with a PATCH attempt counter and now assert that exactly one PATCH request is made. The companion test still covers the multi-attempt fallback path where the first payload shape is rejected.

jacobweinstock (approved):
Thanks for this, @nuxster !

PR #439, opened 2026-06-02
redfishwrapper: return true on successful power off

abbb03:

SystemPowerOff in redfishwrapper returns false, nil even if power off went success. Because of this, bmc/power.go always returns an error. This change corrects this behavior.

### Checklist
- [ ] Tests added

N/A — no dependency changes

```
SystemPowerOff in redfishwrapper now doesn't causes error on successful execution.
```


jacobweinstock (approved):
Hey @abbb03 , good catch. Thanks for this!

PR #443, opened 2026-06-19
fix: add EthernetInterfaces fallback for NICs missing from NetworkInterfaces

mcanevet:

Adds a fallback in `collectNICs` that picks up EthernetInterfaces entries whose MAC addresses were not discovered via the NetworkInterfaces hierarchy. This fixes Supermicro AOC add-on cards, which appear in `EthernetInterfaces` but are absent from `NetworkInterfaces`.

### Checklist
- [x] Tests added

Supermicro

AOC-S25G-m2S (25GbE add-on card). Likely affects other Supermicro AOC add-on cards as well.

Unknown — observed on a production system but firmware version was not captured.

N/A

```
fix: Supermicro AOC add-on NIC cards (e.g. AOC-S25G-m2S) now appear in inventory.
These adapters are absent from NetworkInterfaces but present in EthernetInterfaces;
bmclib now falls back to EthernetInterfaces for any MAC not already discovered.
```

### DIFF
--- a/internal/redfishwrapper/inventory_collect.go
+++ b/internal/redfishwrapper/inventory_collect.go
@@ -127,6 +127,10 @@ func (c *Client) collectNICs(sys *schemas.ComputerSystem, device *common.Device,
 		return err
 	}
 
+	// Fallback: add any EthernetInterfaces not discovered via NetworkInterfaces.
+	// Some adapters (e.g. Supermicro AOC add-on cards) only appear in EthernetInterfaces.
+	discoveredMACs := make(map[string]struct{})
+
 	for _, nic := range nics {
 		// collect network interface adaptor information
 		adapter, err := nic.NetworkAdapter()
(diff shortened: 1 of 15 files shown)


internal/redfishwrapper/inventory_collect.go
```diff
@@ -185,6 +185,55 @@ func (c *Client) collectNICs(sys *schemas.ComputerSystem, device *common.Device,
 		device.NICs = append(device.NICs, n)
 	}
 
+	// Fallback: add any EthernetInterfaces not discovered via NetworkInterfaces.
+	// Some adapters (e.g. Supermicro AOC add-on cards) only appear in EthernetInterfaces.
+	discoveredMACs := make(map[string]struct{})
```
joelrebel: Could this map be populated in the loop above instead?

mcanevet: You're right, I'll do that instead.

mcanevet: fixed

joelrebel (approved):
Thanks for the fix and applying the suggestions

PR #448, opened 2026-07-07
providers/ipmi: fix session lifecycle to use persistent connection

jacobweinstock:

The ipmi provider was opening a new RMCP+ session per operation via an internal ensureConnected/defer Close pattern in every goipmi method. go-ipmi assigns per-session request sequence numbers during the RMCP+ handshake; tearing down the session between calls caused the BMC to reject subsequent requests with IPMI LAN: no matching IPMI response (expected rqSeq 0x05, command 0x38) before UDP read deadline after 5 attempt(s). Chained operations (e.g. PowerSet → IsOn → ChassisControl) compounded this: each inner call's defer Close tore down the session the outer call was still using, causing cascading timeouts and a 20s test runtime. The provider now maintains a single persistent session for the duration of an Open/Close lifecycle, with a mutex-guarded open bool to prevent the double-connect (silent session state overwrite) and double-close (close of closed channel panic) that go-ipmi does not protect against itself.

### Checklist
- [ ] Tests added

```
```

### DIFF
--- a/internal/goipmi/goipmi.go
+++ b/internal/goipmi/goipmi.go
@@ -94,19 +94,16 @@ func toCipherSuiteID(c int) ipmi.CipherSuiteID {
 	return ipmi.CipherSuiteID3
 }
 
-// ensureConnected ensures the IPMI client is connected
-func (i *Ipmi) ensureConnected(ctx context.Context) error {
-	// For go-ipmi, we need to connect before each operation
+func (i *Ipmi) Open(ctx context.Context) error {
 	return i.client.Connect(ctx)
+}
+
+func (i *Ipmi) Close(ctx context.Context) error {
+	return i.client.Close(ctx)
 }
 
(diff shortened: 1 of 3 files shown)


jacobweinstock:
@joelrebel , @appkins , I tried the go-ipmi provider on main and it didnt work at all. did either of you have this working?

joelrebel:
@jacobweinstock haven't had a chance to test this provider as yet, thanks for the review and fixing this up.
I intend to test this out soon on some HP, Dell and SMC hardware in the next weeks.

Should bmclib have a separate registry for 'beta' drivers that aren't enabled by default?

joelrebel (approved):

jacobweinstock:
Thanks, @joelrebel . A separate registry for beta drivers is interesting. It could be useful to show the maturity level and real world use.

PR #451, opened 2026-07-09
fix: don't gate collectDrives on Storage.DrivesCount

mcanevet:

`collectDrives` skipped a `Storage` member entirely when `DrivesCount`
(`Drives@odata.count`) was `0`, even if the `Drives` link array itself was
populated with real, linked drives. `DrivesCount` comes from an *optional*
Redfish property — some implementations populate the `Drives` array without
ever including that count property, so it parses to `0` even though real
drives are linked. The old code treated that as "no drives" and **silently
dropped every drive on that Storage member from the inventory**.

`member.Drives()` reads the underlying link list directly and already
returns an empty, error-free slice when there's genuinely nothing linked, so
this drops the `DrivesCount` gate entirely rather than short-circuiting on a
field that may simply be unset.

### Checklist
- [x] Tests added

Supermicro (observed), but this is a spec-level field misuse — likely
affects any Redfish implementation that links drives without also
populating the optional count property.

Observed on a Supermicro AS-1114S-WN10RT-EU with two KIOXIA KCD8XVUG6T40
NVMe SSDs — both were missing from inventory before this fix.

N/A

```
fix: drives are no longer silently missing from inventory when a Storage
resource links real drives but omits the optional Drives@odata.count
property. collectDrives no longer gates on DrivesCount, which is only
sometimes populated.
```

### DIFF
--- a/internal/redfishwrapper/fixtures/smc_nvme_storage/drive_bay2.json
+++ b/internal/redfishwrapper/fixtures/smc_nvme_storage/drive_bay2.json
@@ -1,0 +1,18 @@
+{
+    "@odata.type": "#Drive.v1_11_1.Drive",
+    "@odata.id": "/redfish/v1/Chassis/NVMeSSD.0.Group.0.StorageBackplane/Drives/Disk.Bay.2",
+    "Name": "Disk.Bay.2",
+    "Id": "Disk.Bay.2",
+    "Manufacturer": "KIOXIA",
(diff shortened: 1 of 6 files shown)


joelrebel (dismissed):
Thanks for fixing this

mcanevet:
@joelrebel rebased

joelrebel (approved):

PR #452, opened 2026-07-09
fix: populate device.Mainboard from Chassis resource data

mcanevet:

`device.Mainboard` was never populated for any vendor — no code path set it,
so it always stayed at `common.NewDevice()`'s empty default.

Some vendors (e.g. Supermicro) don't expose a dedicated motherboard
resource: on a real Supermicro AS-1114S-WN10RT-EU, `Chassis.Model` is the
board's own model (`H12SSW-NTR`) while `Chassis.PartNumber`/`SerialNumber`
describe the case, and the board's own serial number is under the
Supermicro OEM extension (`Oem.Supermicro.BoardSerialNumber`), distinct
from the case serial.

Adds `collectMainboard`, wired alongside the existing `collectEnclosure`, to
populate `device.Mainboard` from these fields. Best-effort: falls back to no
serial number on chassis without the Supermicro OEM extension, and does
nothing on chassis without a board model at all.

`chassisAttributes` calls this once per compatible `Chassis` member, and
some vendors (e.g. Dell) have more than one compatible Chassis ID in
`KnownChassisOdataIDs`. Since `device.Mainboard` is a single field, not a
slice, `collectMainboard` only populates it from the first chassis that
provides a model — this prevents an unrelated chassis processed later (e.g.
a drive enclosure/backplane with its own `Model` set) from silently
overwriting a correct motherboard entry.

### Checklist
- [x] Tests added

Supermicro (data source), but the "don't overwrite" guard is written
generically for any vendor with multiple compatible Chassis resources
(e.g. Dell).

Observed on a Supermicro AS-1114S-WN10RT-EU (motherboard: H12SSW-NTR).

N/A

```
feat: populate device.Mainboard (vendor, model, serial) from the Chassis
resource, previously always empty for every vendor. On Supermicro,
extracts the board's own serial from the vendor OEM extension, distinct
from the case serial.
```

### DIFF
--- a/internal/redfishwrapper/inventory_collect.go
+++ b/internal/redfishwrapper/inventory_collect.go
@@ -42,6 +42,46 @@ func (c *Client) collectEnclosure(ch *schemas.Chassis, device *common.Device, so
 	return nil
 }
 
+// collectMainboard collects motherboard information.
+//
+// Some vendors (e.g. Supermicro) don't expose a dedicated motherboard
+// resource: the Chassis resource's Model is the board's own model, while
+// PartNumber/SerialNumber describe the case rather than the board itself.
+// Vendor-specific providers may fill in additional Mainboard attributes
+// (e.g. the board's own serial number) after Inventory returns.
+//
+// chassisAttributes calls this once per compatible Chassis member, and some
+// vendors (e.g. Dell) have more than one compatible Chassis ID. device.Mainboard
+// is a single field, not a slice, so only populate it from the first chassis
+// that provides a model — don't let an unrelated chassis (e.g. a drive
+// enclosure/backplane with its own Model set) silently overwrite it later.
+func (c *Client) collectMainboard(ch *schemas.Chassis, device *common.Device, softwareInventory []*schemas.SoftwareInventory) (err error) {
+	if ch == nil || ch.Model == "" || (device.Mainboard != nil && device.Mainboard.Model != "") {
+		return nil
+	}
+
+	m := &common.Mainboard{
+		Common: common.Common{
+			Vendor: common.FormatVendorName(ch.Manufacturer),
+			Model:  ch.Model,
+			Status: &common.Status{
+				Health: string(ch.Status.Health),
+				State:  string(ch.Status.State),
+			},
+			Firmware: &common.Firmware{},
+		},
+
+		PhysicalID: ch.ID,
+	}
+
+	// include additional firmware attributes from redfish firmware inventory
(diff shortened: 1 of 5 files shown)


internal/redfishwrapper/inventory_collect.go
```diff
@@ -42,6 +43,67 @@ func (c *Client) collectEnclosure(ch *schemas.Chassis, device *common.Device, so
...
+func (c *Client) collectMainboard(ch *schemas.Chassis, device *common.Device, softwareInventory []*schemas.SoftwareInventory) (err error) {
+	if ch == nil || ch.Model == "" || (device.Mainboard != nil && device.Mainboard.Model != "") {
+		return nil
+	}
+
+	m := &common.Mainboard{
+		Common: common.Common{
+			Vendor: common.FormatVendorName(ch.Manufacturer),
+			Model:  ch.Model,
+			Serial: supermicroBoardSerialNumber(ch.OEM),
```
joelrebel: We'd want to keep the redfishwrapper Inventory method to be vendor neutral, 
the vendor specific changes can go into the Supermicro specific `Inventory` method - https://github.com/bmc-toolbox/bmclib/blob/main/providers/supermicro/supermicro.go#L265

mcanevet:
@joelrebel done

providers/redfish/redfish.go
```diff
@@ -229,7 +230,46 @@ func (c *Conn) SetVirtualMedia(ctx context.Context, kind, mediaURL string) (ok b
...
+	populateMainboardSerial(ctx, c.redfishwrapper, device)
+
+	return device, nil
+}
+
+// populateMainboardSerial fills in device.Mainboard.Serial from the Chassis
+// resource's Supermicro OEM extension: Chassis.SerialNumber is the case
+// serial, while the board's own serial is only available under
+// Oem.Supermicro.BoardSerialNumber.
+func populateMainboardSerial(ctx context.Context, rf *redfishwrapper.Client, device *common.Device) {
```
joelrebel: This method needs to be removed from here as its now in the Supermicro provider

joelrebel (changes requested):

providers/redfish/redfish.go
mcanevet: Removed — `redfish.go`'s `Inventory` now just passes through to `redfishwrapper.Inventory` again, no vendor-specific enrichment. Pushed in 712fcab.

joelrebel (approved):

PR #453, opened 2026-07-09
fix: populate device.BIOS.Vendor from system manufacturer

mcanevet:

`collectBIOS` never set `BIOS.Vendor`, so it stayed empty for every vendor.
Sets it from the `ComputerSystem`'s own `Manufacturer` field, the same
source already used elsewhere in this file for other components' vendor
(e.g. `collectEnclosure`, `collectPSUs`).

### Checklist
- [x] Tests added

Supermicro (observed), but this affects every vendor — `BIOS.Vendor` was
unconditionally empty regardless of BMC make.

Observed on a Supermicro AS-1114S-WN10RT-EU.

N/A

```
fix: BIOS.Vendor is now populated from the system manufacturer. It was
previously always empty regardless of vendor.
```


joelrebel (approved):
Thanks for fixing this

PR #456, opened 2026-08-06
fix: split BiosConfigurationSetter into map and file interfaces

mcanevet:

`BiosConfigurationSetter` currently requires implementing **both** `SetBiosConfiguration` and `SetBiosConfigurationFromFile`. `providers/redfish`, `providers/dell/idrac`, and `providers/lenovo` only implement `SetBiosConfiguration` - the file-import path is a Dell/Supermicro-SUM-specific whole-config-blob import with no generic Redfish equivalent - so none of them satisfy the interface today. `Client.SetBiosConfiguration()` fails with `no BiosConfigurationSetter implementations found` on those providers even though the underlying call works fine.

Confirmed live against a real Supermicro AS-1114S-WN10RT-EU BMC: on current `main`, `SetBiosConfiguration` fails with `not a BiosConfigurationSetter implementation: *redfish.Conn` (and `*lenovo.Conn`) before any network request is even made; after this fix, dispatch correctly selects those providers and the request goes out.

Splits `BiosConfigurationSetter` (map only) from a new `BiosConfigurationFileSetter` (file only), dispatched independently. `providers/supermicro`, which implements both, is unaffected.

Also adds `var _ Interface = (*Conn)(nil)` compile-time assertions (the pattern already used in `providers/lenovo`'s `bmc.go`/`virtual_media.go`/`sel.go`/etc., but never applied to the BIOS interfaces) to `redfish`, `dell/idrac`, `lenovo`, and `supermicro`, so this exact class of bug can't regress silently again - applying the same assertions to unmodified `main` fails to compile in exactly the three affected providers.

### Checklist
- [x] Tests added

Any BMC reached via the generic Redfish, Dell iDRAC, or Lenovo XClarity providers. Confirmed live against Supermicro.

Confirmed live against a Supermicro AS-1114S-WN10RT-EU.

N/A - this is a Go interface-satisfaction bug, not firmware-dependent.

None.

```
Fix BiosConfigurationSetter so the redfish, dell/idrac, and lenovo providers are recognized as implementing SetBiosConfiguration.
```

### DIFF
--- a/bmc/bios.go
+++ b/bmc/bios.go
@@ -20,16 +20,30 @@ type biosConfigurationGetterProvider struct {
 	BiosConfigurationGetter
 }
 
-// BiosConfigurationSetter provides applying a host's BIOS configuration,
-// either from a map of settings or from a configuration file.
+// BiosConfigurationSetter provides applying a host's BIOS configuration from a
+// map of settings.
 type BiosConfigurationSetter interface {
 	SetBiosConfiguration(ctx context.Context, biosConfig map[string]string) (err error)
-	SetBiosConfigurationFromFile(ctx context.Context, cfg string) (err error)
 }
 
 type biosConfigurationSetterProvider struct {
 	name string
 	BiosConfigurationSetter
+}
+
+// BiosConfigurationFileSetter provides applying a host's BIOS configuration from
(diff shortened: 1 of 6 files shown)


joelrebel (approved):
Thanks, splitting this out is ideal

PR #457, opened 2026-08-06
feat: add UEFI SecureBoot enable/disable and key reset support

mcanevet:

bmclib has no way to enable/disable UEFI SecureBoot or reset its key databases, despite gofish (already a direct dependency) fully supporting both via the ComputerSystem's SecureBoot resource (`SecureBootEnable` + `.Update()`, and `.ResetKeys()`).

Adds the standard three-layer bmclib capability, following the exact pattern of `bmc/bios.go`/`bmc/boot_device.go`:
- `internal/redfishwrapper`: `GetSecureBoot`/`SetSecureBoot`/`ResetSecureBootKeys` wrapping gofish's `schemas.SecureBoot`.
- `bmc`: `SecureBootStateGetter`/`SecureBootSetter`/`SecureBootKeysResetter` interfaces with the usual `FromInterfaces` dispatch.
- `Client.GetSecureBoot`/`SetSecureBoot`/`ResetSecureBootKeys` passthroughs.

Wired into every provider that embeds `redfishwrapper.Client` - `redfish` (generic), `dell`, `supermicro`, and `lenovo` - plus new `Feature*` constants (`FeatureGetSecureBoot`, `FeatureSetSecureBoot`, `FeatureResetSecureBootKeys`) on each. This matches how every other redfishwrapper-backed capability (BIOS config, power state, NMI, ...) is exposed at each provider layer, even though the implementation is the same delegation in every case.

### Checklist
- [x] Tests added

Any BMC reached via the generic Redfish, Dell, Supermicro, or Lenovo XCC provider.

N/A - generic Redfish SecureBoot resource, not vendor-specific.

Requires the standard Redfish `SecureBoot` resource under `ComputerSystem`.

gofish (already a direct dependency, `github.com/stmcginnis/gofish/schemas`).

```
Add SetSecureBoot/GetSecureBoot/ResetSecureBootKeys support for the redfish, dell, supermicro, and lenovo providers.
```

### DIFF
--- a/internal/redfishwrapper/secure_boot.go
+++ b/internal/redfishwrapper/secure_boot.go
@@ -4,6 +4,8 @@ import (
 	"context"
 
 	"github.com/stmcginnis/gofish/schemas"
+
+	bmclibErrs "github.com/bmc-toolbox/bmclib/v2/errors"
 )
 
 // GetSecureBoot returns whether UEFI Secure Boot is currently enabled for the system.
@@ -14,7 +16,7 @@ func (c *Client) GetSecureBoot(ctx context.Context) (enabled bool, err error) {
 	}
 
 	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
-		return false, nil
+		return false, bmclibErrs.ErrRedfishSystemOdataID
 	}
 
 	secureBoot, err := sys.SecureBoot()
@@ -34,7 +36,7 @@ func (c *Client) SetSecureBoot(ctx context.Context, enable bool) (err error) {
 	}
 
 	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
-		return nil
+		return bmclibErrs.ErrRedfishSystemOdataID
 	}
 
 	secureBoot, err := sys.SecureBoot()
@@ -56,7 +58,7 @@ func (c *Client) ResetSecureBootKeys(ctx context.Context, resetType string) (err
 	}
 
 	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
-		return nil
+		return bmclibErrs.ErrRedfishSystemOdataID
 	}
 
 	secureBoot, err := sys.SecureBoot()
--- a/internal/redfishwrapper/secure_boot.go
+++ b/internal/redfishwrapper/secure_boot.go
@@ -1,0 +1,70 @@
+package redfishwrapper
+
+import (
+	"context"
+
+	"github.com/stmcginnis/gofish/schemas"
+)
+
+// GetSecureBoot returns whether UEFI Secure Boot is currently enabled for the system.
+func (c *Client) GetSecureBoot(ctx context.Context) (enabled bool, err error) {
+	sys, err := c.System()
+	if err != nil {
+		return false, err
+	}
+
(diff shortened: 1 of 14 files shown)


internal/redfishwrapper/secure_boot.go
```diff
@@ -0,0 +1,70 @@
...
+
+// GetSecureBoot returns whether UEFI Secure Boot is currently enabled for the system.
+func (c *Client) GetSecureBoot(ctx context.Context) (enabled bool, err error) {
+	sys, err := c.System()
+	if err != nil {
+		return false, err
+	}
+
+	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
+		return false, nil
```
joelrebel: This should return an error indicating the incompatibility, 
without that the wrapper client https://github.com/bmc-toolbox/bmclib/pull/457/changes/66980e7a2c0bc9d0483a32b501617ed48fd27a24 will not continue attempting another provider if one of them returns false.

internal/redfishwrapper/secure_boot.go
```diff
@@ -0,0 +1,70 @@
...
+// ResetSecureBootKeys resets the UEFI Secure Boot key databases. resetType is one
+// of ResetAllKeysToDefault, DeleteAllKeys, or DeletePK.
+func (c *Client) ResetSecureBootKeys(ctx context.Context, resetType string) (err error) {
+	sys, err := c.System()
+	if err != nil {
+		return err
+	}
+
+	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
+		return nil
```
joelrebel: If its not compatible, this should indicate an error so that the next provider is attempted by the client wrapper

internal/redfishwrapper/secure_boot.go
```diff
@@ -0,0 +1,70 @@
...
+// SetSecureBoot enables or disables UEFI Secure Boot for the system. The system
+// must be in UEFI boot mode for this to take effect.
+func (c *Client) SetSecureBoot(ctx context.Context, enable bool) (err error) {
+	sys, err := c.System()
+	if err != nil {
+		return err
+	}
+
+	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
+		return nil
```
joelrebel: Similar to comment above, return error when not compatible or consider if this check provides much value.

joelrebel (changes requested):

internal/redfishwrapper/secure_boot.go
mcanevet: Good catch, fixed in 7ef2e64 — now returns `ErrRedfishSystemOdataID` instead of `false, nil`.

internal/redfishwrapper/secure_boot.go
mcanevet: Fixed in 7ef2e64 — returns `ErrRedfishSystemOdataID` now.

internal/redfishwrapper/secure_boot.go
mcanevet: Fixed in 7ef2e64 — returns `ErrRedfishSystemOdataID` now, matching `inventory.go`'s handling of the same check.

joelrebel (approved):

PR #459, opened 2026-08-06
redfishwrapper: verify boot device set actually took effect

mcanevet:
## What

`SystemBootDeviceSet` returned `ok=true` as soon as the `PATCH` got a
`200 OK`, without checking whether the BMC actually stored what was
requested.

## Why

On a real Supermicro `AS-1015CS-TNR-EU` (AMI-based Redfish,
`RedfishVersion: 1.22.2`), a `PATCH` requesting a one-time UEFI PXE
override (`BootSourceOverrideTarget=Pxe`,
`BootSourceOverrideEnabled=Once`, `BootSourceOverrideMode=UEFI`) got a
`200 OK`, but a subsequent `GET` showed `BootSourceOverrideMode`
silently reverted to `"Legacy"` and `Enabled` to `"Disabled"` - only
`Target` stuck. This system also has CSM/Legacy boot disabled in its
BIOS, so the override could never actually be honored, but callers had
no way to detect this short of manually re-checking after the call
returned.

We hit this in production via Tinkerbell's rufio controller, which
uses `SystemBootDeviceSet` to set a one-time PXE boot override before
provisioning bare-metal hosts - the box never actually rebooted into
PXE despite bmclib reporting success.

## What changed

`SystemBootDeviceSet` now re-fetches the system after writing and
compares against what was actually sent - not necessarily the full
request, since the existing minimal-fields fallback (for Redfish
implementations that reject the full field set, e.g. HP DL160 Gen10)
deliberately omits `Mode`, so verification only checks fields that
were actually part of whichever request succeeded. Returns `ok=false`
with a descriptive error on mismatch instead of a false `ok=true`.

Fixtures in `internal/redfishwrapper/fixtures/supermicro/` are real
captures from the affected hardware (see that directory's `README.md`)
plus one synthetic fixture for the added happy-path regression test,
confirming a BMC that behaves correctly still gets `ok=true`.

## Testing

- `go test ./...` passes across the whole repo.
- New test `TestSystemBootDeviceSet_DetectsSupermicroSilentModeDowngrade`
  reproduces the real bug via fixtures and confirms it's now caught.
- New test `TestSystemBootDeviceSet_HappyPath` confirms no regression
  for BMCs that honor the request correctly.

### DIFF
--- a/internal/redfishwrapper/boot_device.go
+++ b/internal/redfishwrapper/boot_device.go
@@ -117,6 +117,7 @@ func (c *Client) SystemBootDeviceSet(_ context.Context, bootDevice string, setPe
 		boot.BootSourceOverrideMode = rf.LegacyBootSourceOverrideMode
 	}
 
+	sent := boot
 	if err = system.SetBoot(boot); err != nil {
 		// Some redfish implementations don't like all the fields we're setting so we
 		// try again here with a minimal set of fields. This has shown to work with the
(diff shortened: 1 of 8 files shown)


mcanevet:
Closing - opened without proper authorization, apologies for the noise.

PR #460, opened 2026-08-26
feat: add per-database UEFI Secure Boot key reset and certificate import

mcanevet:

Extends the SecureBoot support added in #457 with per-database operations, newly possible now that gofish's `SecureBootDatabase` gained `AddCertificate`/`AddSignature` (stmcginnis/gofish#561). Unlike the whole-subsystem `ResetSecureBootKeys`, these target a single database (e.g. `db`, `KEK`) without touching `PK` - the mechanism a caller needs to enroll a custom certificate into `db`/`KEK` while leaving an already-provisioned `PK` (and whatever Setup Mode state it implies) alone.

Adds the same three-layer bmclib capability as #457, wired into the same four providers:
- `internal/redfishwrapper`: `ResetSecureBootDatabaseKeys`/`ImportSecureBootCertificate`, finding the target database by `DatabaseID` in the `SecureBootDatabases` collection.
- `bmc`: `SecureBootDatabaseKeysResetter`/`SecureBootCertificateImporter` interfaces with the usual `FromInterfaces` dispatch.
- `Client.ResetSecureBootDatabaseKeys`/`ImportSecureBootCertificate` passthroughs.

Wired into every provider that embeds `redfishwrapper.Client` - `redfish` (generic), `dell`, `supermicro`, and `lenovo` - plus new `Feature*` constants (`FeatureResetSecureBootDatabaseKeys`, `FeatureImportSecureBootCertificate`) on each, matching how `GetSecureBoot`/`SetSecureBoot`/`ResetSecureBootKeys` are already exposed at each provider layer.

Bumps gofish to a pseudo-version pinned to the `AddCertificate`/`AddSignature` commit (`stmcginnis/gofish@aa6a0d7`), ahead of the next tagged release - happy to hold this until a real gofish tag exists if you'd prefer.

### Checklist
- [x] Tests added

Any BMC reached via the generic Redfish, Dell, Supermicro, or Lenovo XCC provider - currently verified end-to-end against real Dell hardware (iDRAC); Supermicro/Lenovo firmware doesn't yet expose `SecureBootDatabases` in our testing, so they'll pick this up automatically once their firmware does.

N/A - generic Redfish `SecureBootDatabase`/`Certificates` resources, not vendor-specific.

Requires the Redfish `SecureBoot.v1_1_x` `SecureBootDatabases` resource collection with a working `Certificates` POST - confirmed present on Dell iDRAC (`SecureBoot.v1_1_2`). Not yet present on the Supermicro/Lenovo firmware we tested against.

gofish `>= aa6a0d7` (`stmcginnis/gofish#561`, `SecureBootDatabase.AddCertificate`/`AddSignature`) - not in a tagged release yet.

```
Add ResetSecureBootDatabaseKeys/ImportSecureBootCertificate support for the redfish, dell, supermicro, and lenovo providers.
```

### DIFF
--- a/bmc/secure_boot.go
+++ b/bmc/secure_boot.go
@@ -10,6 +10,36 @@ import (
 	bmclibErrs "github.com/bmc-toolbox/bmclib/v2/errors"
 )
 
+// ResetSecureBootKeysType identifies the type of Secure Boot key reset to perform.
+type ResetSecureBootKeysType string
+
+// ResetSecureBootKeysType constants enumerate the supported reset types for the whole Secure Boot subsystem.
+const (
+	ResetSecureBootKeysTypeResetAllKeysToDefault ResetSecureBootKeysType = "ResetAllKeysToDefault"
+	ResetSecureBootKeysTypeDeleteAllKeys         ResetSecureBootKeysType = "DeleteAllKeys"
+	ResetSecureBootKeysTypeDeletePK              ResetSecureBootKeysType = "DeletePK"
+)
+
+// ResetSecureBootDatabaseKeysType identifies the type of per-database Secure Boot key reset to perform.
+type ResetSecureBootDatabaseKeysType string
+
+// ResetSecureBootDatabaseKeysType constants enumerate the supported reset types for a single Secure Boot database.
+const (
+	ResetSecureBootDatabaseKeysTypeResetAllKeysToDefault ResetSecureBootDatabaseKeysType = "ResetAllKeysToDefault"
+	ResetSecureBootDatabaseKeysTypeDeleteAllKeys         ResetSecureBootDatabaseKeysType = "DeleteAllKeys"
+)
+
+// SecureBootDatabase identifies a UEFI Secure Boot key database.
+type SecureBootDatabase string
+
+// SecureBootDatabase constants enumerate the supported Secure Boot key databases.
+const (
+	SecureBootDatabaseDB  SecureBootDatabase = "db"
+	SecureBootDatabaseKEK SecureBootDatabase = "KEK"
+	SecureBootDatabasePK  SecureBootDatabase = "PK"
+	SecureBootDatabaseDBX SecureBootDatabase = "dbx"
+)
+
 // SecureBootStateGetter provides retrieval of whether UEFI Secure Boot is enabled.
 type SecureBootStateGetter interface {
 	GetSecureBoot(ctx context.Context) (enabled bool, err error)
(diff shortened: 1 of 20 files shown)


bmc/secure_boot.go
```diff
@@ -40,6 +40,30 @@ type secureBootKeysResetterProvider struct {
 	SecureBootKeysResetter
 }
 
+// SecureBootDatabaseKeysResetter provides resetting a single UEFI Secure Boot key
+// database (e.g. "db", "KEK"), leaving every other database - including PK -
+// untouched.
+type SecureBootDatabaseKeysResetter interface {
+	ResetSecureBootDatabaseKeys(ctx context.Context, database, resetType string) (err error)
```
joelrebel: I should have raised this in the first PR for secureboot support, can we change the `database`, `resetType` fields to be typed constants and not plain strings. That makes it clear to users of this library what can be input here and puts less burden for the library to validate input.

internal/redfishwrapper/secure_boot.go
```diff
@@ -70,3 +70,65 @@ func (c *Client) ResetSecureBootKeys(ctx context.Context, resetType string) (err
...
+	sys, err := c.System()
+	if err != nil {
+		return nil, err
+	}
+
+	if !c.compatibleOdataID(sys.ODataID, knownSystemsOdataIDs) {
+		return nil, bmclibErrs.ErrRedfishSystemOdataID
+	}
+
+	secureBoot, err := sys.SecureBoot()
```
joelrebel, line 95: The Gofish `SecureBoot()` method returns `nil`, `nil` if the resource link is empty on the BMC,
we need to have a nil check added for `secureBoot`
ref: https://github.com/stmcginnis/gofish/blob/main/schemas/computersystem.go#L1539

Please also add the nil check to the other methods in this file where the `SecureBoot()` method from Gofish is being invoked.

joelrebel (changes requested):
@mcanevet thanks for this PR,
Can you include some examples for Secure Boot use in the https://github.com/bmc-toolbox/bmclib/tree/main/examples 

The gofish pseudo version pin is fine, once that repo has a tagged release, it can be updated here. 

Going ahead, please try to split the provider changes from the internal/redfishwrapper commits, that gives us the chance to revert specific provider changes easier.

mcanevet:
@joelrebel done

mcanevet:
I forgot to split per provider. Now it's OK

joelrebel (approved):
Thanks for this contribution and addressing the feedback 👍

PR #463, opened 2026-09-08
feat: add SetHTTPBootURI support

mcanevet:
Adds a new `HTTPBootURISetter` interface/dispatcher for setting the URI
UEFI HTTP Boot fetches its boot image from, wired for the generic Redfish,
Dell, and Supermicro providers.

- `bmc.SetHTTPBootURIFromInterfaces` / `Client.SetHTTPBootURI`: new
  dispatcher, following the existing SetVirtualMedia pattern.
- `internal/redfishwrapper`: PATCHes the standard Redfish
  ComputerSystem.Boot.HttpBootUri property (schema v1.9.0+). Setting the
  URI does not select UEFI HTTP Boot as the boot target -
  BootSourceOverrideTarget/Enabled/Mode are left untouched; use
  SetBootDevice(..., "uefi_http", ...) for that.
- `providers/redfish`, `providers/supermicro`: delegate to the above -
  both vendors expose the standard property.
- `providers/dell`: Dell's iDRAC does not populate
  ComputerSystem.Boot.HttpBootUri at all (confirmed live, absent from the
  Boot object entirely, not just empty, on iDRAC firmware 1.5.3). Dell
  instead models the URI as a per-NIC BIOS Setup attribute (HttpDevNUri),
  so this PATCHes that attribute via SetBiosConfiguration instead,
  targeting device slot 1 (there's no field yet to select a NIC).

Independent of the HTTP Boot enable/disable toggle (see the companion
SetNetworkBootEnabled PR) - setting the URI doesn't enable the capability,
and enabling the capability doesn't require a URI.

Tested live end-to-end via Tinkerbell workflows: Dell PowerEdge R6715
(iDRAC 1.5.3) and a Supermicro H12SSW-NTR board, both successfully
UEFI-HTTP-booting an iPXE binary using the URI this sets.

### DIFF
--- a/client.go
+++ b/client.go
@@ -592,6 +592,18 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind, mediaURL string) (ok
 	return ok, err
 }
 
+// SetHTTPBootURI sets the URI UEFI HTTP Boot fetches its boot image from.
+func (c *Client) SetHTTPBootURI(ctx context.Context, uri string) (ok bool, err error) {
+	ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "SetHTTPBootURI")
+	defer span.End()
+
+	ok, metadata, err := bmc.SetHTTPBootURIFromInterfaces(ctx, uri, c.registry().GetDriverInterfaces())
+	c.setMetadata(metadata)
+	metadata.RegisterSpanAttributes(c.Auth.Host, span)
+
+	return ok, err
+}
+
 // ResetBMC pass through to library function
 func (c *Client) ResetBMC(ctx context.Context, resetType string) (ok bool, err error) {
 	ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "ResetBMC")
--- a/bmc/http_boot.go
+++ b/bmc/http_boot.go
@@ -1,0 +1,74 @@
+package bmc
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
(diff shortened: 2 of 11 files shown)


client.go
```diff
@@ -592,6 +592,18 @@ func (c *Client) SetVirtualMedia(ctx context.Context, kind, mediaURL string) (ok
 	return ok, err
 }
 
+// SetHTTPBootURI sets the URI UEFI HTTP Boot fetches its boot image from.
+func (c *Client) SetHTTPBootURI(ctx context.Context, uri string) (ok bool, err error) {
+	ctx, span := c.traceprovider.Tracer(pkgName).Start(ctx, "SetHTTPBootURI")
+	defer span.End()
+
```
joelrebel, line 599: Would it be worth checking the `uri` string is not empty before proceeding here?

mcanevet: Empty string is meant to be how you unset it, not something to reject. Right now it doesn't actually work on the generic Redfish/Supermicro path though - gofish's `Boot.HTTPBootURI` has `omitempty` but isn't a pointer, so an empty string gets dropped from the PATCH body instead of sent, and the BMC never sees the clear request. Filed a fix upstream: stmcginnis/gofish#569 (unmerged). Dell's path (BIOS attribute, not this property) does send the empty value mechanically, but I haven't confirmed live that iDRAC accepts it.

So: no check here for now - once #569 lands we can bump gofish and it'll work as intended.

joelrebel (approved):

PR #464, opened 2026-09-08
feat: add SetNetworkBootEnabled support

mcanevet:
Adds a new `NetworkBootEnabledSetter` interface/dispatcher for
enabling/disabling UEFI HTTP Boot and/or legacy PXE boot capability in
BIOS/UEFI firmware - independent of SetHTTPBootURI (see the companion PR):
enabling/disabling the capability and pointing it at a boot image are
separate concerns with no reason to be coupled.

- `bmc.SetNetworkBootEnabledFromInterfaces` / `Client.SetNetworkBootEnabled`:
  new dispatcher. httpEnabled/pxeEnabled are independent - a nil pointer
  leaves that protocol untouched, so either, both, or neither may be set.
- `internal/redfishwrapper`: carries a BIOS-attribute fingerprint table
  (currently one entry: Supermicro/AMI Aptio, confirmed live on an
  AS-1114S-WN10RT-EU) that resolves the right PATCH body from a machine's
  current GetBiosConfiguration result. This previously lived in a
  downstream consumer (tinkerbell/rufio) - per-vendor BIOS attribute
  knowledge belongs in bmclib so any caller gets it, not just the one
  that happened to need it first. Returns a clear error rather than
  silently no-op'ing on an unrecognized BIOS.
- `providers/redfish`, `providers/supermicro`: delegate to the above.
- `providers/dell`: targets Dell's per-NIC HttpDevNEnDis/PxeDevNEnDis BIOS
  Setup attributes instead, which the generic fingerprint table doesn't
  recognize - device slot 1, same as the companion SetHTTPBootURI PR.

Tested live end-to-end via Tinkerbell workflows on the same two boards as
the companion PR - the fingerprint table applies and takes effect
(confirmed via a real reboot and UEFI HTTP Boot attempt), and the Dell
attributes were confirmed via a live PATCH against a PowerEdge R6715.

### DIFF
--- a/internal/redfishwrapper/network_boot.go
+++ b/internal/redfishwrapper/network_boot.go
@@ -1,0 +1,117 @@
+package redfishwrapper
+
+import (
+	"context"
+	"fmt"
+)
+
+const (
+	attrEnabled  = "Enabled"
+	attrDisabled = "Disabled"
+	attrUEFI     = "UEFI"
+)
+
+// networkBootFingerprintTables resolves the BIOS attribute PATCH body needed to enable or
+// disable UEFI HTTP Boot and/or legacy PXE boot capability, based on a fingerprint match against
+// a machine's current BIOS configuration.
+//
+// fingerprint keys are attribute names that, if present in a GetBiosConfiguration() result,
+// identify this table as applicable. Extend with one entry per additional vendor/board as
+// fleets need them — do not attempt to support unknown BIOS/vendor combinations silently.
+//
+// httpEnabled/httpDisabled and pxeEnabled/pxeDisabled are independent attribute sets: HTTP Boot
+// and legacy PXE are separate capabilities that can be on or off in any combination. NetworkStack
+// and BootModeSelect are prerequisites shared by both protocols, so they're only asserted on the
+// enabled side of each set — disabling one protocol must not turn off the network stack the
+// other protocol may still depend on.
+var networkBootFingerprintTables = []struct {
+	fingerprint  string // attribute name unique enough to identify this BIOS/vendor
+	httpEnabled  map[string]string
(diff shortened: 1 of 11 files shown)


internal/redfishwrapper/network_boot.go
```diff
@@ -0,0 +1,117 @@
...
+// other protocol may still depend on.
+var networkBootFingerprintTables = []struct {
+	fingerprint  string // attribute name unique enough to identify this BIOS/vendor
+	httpEnabled  map[string]string
+	httpDisabled map[string]string
+	pxeEnabled   map[string]string
+	pxeDisabled  map[string]string
+}{
+	{
+		fingerprint: "IPv4HTTPSupport", // Supermicro H12SSW-NTR / AMI Aptio, confirmed io14-72
```
joelrebel: Would it be better if this was moved within the provider, and if the fingerprint does not match, clearly indicate that to the caller?

mcanevet: I think this is meant to stay here rather than move - the same board can be reached either via `providers/supermicro` or the generic `providers/redfish` driver (bmclib tries both by default), and fingerprinting off the live BIOS config in the shared layer means either path picks up the right attributes. Moving it into `providers/supermicro` would probably drop that for the generic-driver path.

Dell's version of this PR hardcodes the attributes directly with no table, since that package already knows the vendor - Supermicro needs the table precisely because this code doesn't know the vendor upfront.

On the second point, I think it already does - `networkBootAttributes` a few lines down returns an error if nothing matches, doesn't no-op. Let me know if you meant something else by that.

joelrebel (approved):
Thanks for the contribution here,
Would suggest to include an example for this new feature within the examples directory in a future PR.

PR #465, opened 2026-09-10
fix(redfish): don't let the connect deadline become the per-request timeout

mcanevet:
## Summary

`Open()` bounds the connect by setting the HTTP client's `Timeout` from the ctx deadline, because gofish stores one context at construction and ignores the context passed to individual calls. That override was never undone, so the value leaked into every subsequent call on the connection.

Callers pass a *connect* deadline, and bmclib divides it further: `Client.defaultTimeout` is `ctx-remaining/len(registry.Drivers)`, so with the 9 providers registered by default a 60s connect budget leaves ~6.6s - which then silently caps every later operation. A Dell `Bios/Settings` PATCH takes ~10s (the iDRAC creates a Lifecycle Controller job before returning 202), so it could never complete within that leftover budget, while the BMC applied the change anyway a moment after the client hung up: a reported failure that actually took effect.

This restores the previous `Timeout` value once the connect returns, so `ctx` bounds `Open` only and later calls use the client's configured timeout.

## Test plan
- Added `TestOpenRestoresHTTPClientTimeout`, verifying the configured `http.Client` timeout is unchanged after `Open` returns.
- Added `TestOpenBoundsConnectByContext`, verifying `Open` still fails around the short ctx deadline rather than hanging until the client's much longer configured timeout.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues), `go test ./...` all pass.

### DIFF
--- a/internal/redfishwrapper/client.go
+++ b/internal/redfishwrapper/client.go
@@ -152,8 +152,26 @@ func (c *Client) Open(ctx context.Context) error {
 		config.DumpWriter = os.Stdout
 	}
 
+	// gofish stores a single context at construction and ignores the context passed to
+	// individual calls, so bounding the connect below by ctx requires setting the HTTP
+	// client's own Timeout. That must not be done on config.HTTPClient itself: gofish.Connect
+	// stores this exact pointer into the resulting APIClient and reuses it for every later call
+	// on this connection, and WithHTTPClient lets a caller share one *http.Client across
+	// multiple bmclib connections - mutating its Timeout in place would race with a concurrent
+	// Open on another connection sharing that same client. Use a private shallow copy instead:
+	// Transport (and so the underlying connection pool) is still shared, but Timeout becomes
+	// independent, and nothing outside this call ever observes it.
+	//
+	// The copy's Timeout is restored before Open returns, since gofish.Connect's stored pointer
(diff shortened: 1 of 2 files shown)


joelrebel (approved):

joelrebel:
@mcanevet, can you confirm if you have tried the PerProviderTimeout option which can be passed in https://github.com/bmc-toolbox/bmclib#timeouts? That should give the Gofish provider a fixed time slice

mcanevet:
Tried it. On the pre-fix code: `PerProviderTimeout=1s` fails a 2s BIOS write at ~1s; bump it to 5s and the same write succeeds. So yes, it works around this specific case - the leaked `HTTPClient.Timeout` from `Open()` and `PerProviderTimeout` are the same number.

Two reasons I'd still rather fix it properly:

It's one knob doing two jobs. Whatever you set it to also becomes how long you wait to notice a dead BMC - set it high enough to survive a slow write and a hung BMC now takes just as long to fail.

And it's not scoped to the connection. `WithHTTPClient` takes a caller-owned client, and nothing stops passing the same one to multiple bmclib connections. I set up two BMCs sharing one client - A opened with a 5s budget expecting its own slow write, B opened after with 1s. A's write then failed at ~1s instead of the 5s it asked for - B's `Open()` had silently overwritten the shared client's timeout out from under it.

PR #466, opened 2026-09-10
feat: add CustomSecureBootKeysAllower for out-of-band custom key acceptance

mcanevet:
## Summary

Adds `bmc.CustomSecureBootKeysAllower`, a new interface controlling whether a platform will accept custom UEFI Secure Boot keys (as opposed to only the vendor-shipped key set). This is distinct from `SecureBootSetter` (enabling/disabling Secure Boot itself) and from `SecureBootCertificateImporter` (enrolling a specific certificate) - it's the out-of-band gate some vendors require before a certificate import will be accepted at all.

Implemented for Dell, which models this as the `SecureBootPolicy` BIOS attribute (`Standard`/`Custom`). Not implemented for Lenovo yet: XCC-based systems have an analogous `SecureBootPolicy` attribute, but it's unconfirmed whether it's reachable through the same generic `/Bios` attribute PATCH this package already uses, or requires Lenovo's proprietary OneCLI/XCC transport - documented as a known gap with links to Lenovo's public docs (including the `FQXSFPU4097G` error code callers will hit without it).

`rebootRequired` reports that a successful change is staged and takes effect only after a power cycle, matching how BIOS Setup attribute changes work generally.

This is self-contained and doesn't depend on any other open PR. There is one known follow-up, which I'll open separately because it only becomes meaningful on top of #468: `AllowCustomSecureBootKeys` currently returns early when the attribute's currently *applied* value already matches the request, which is unsafe if a *different* value is genuinely pending from an earlier call in the same boot cycle. On `main` today that early return is behaviourally identical to removing it (gofish's `UpdateBiosAttributesApplyAt` diffs the attribute away and sends nothing either way), so it isn't a live regression here - it only becomes both reachable and fixable once #468 lands.

## Test plan
- `bmc/secure_boot_test.go`: `TestAllowCustomSecureBootKeysFromInterfaces` covering success/no-implementation/error cases.
- `providers/dell/secure_boot_test.go`: enable-from-Standard, already-Custom (no-op), disable, and the attribute-absent (`ErrUnsupportedHardware`) case.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 new issues), `go test ./...` all pass.


bmc/secure_boot.go
```diff
@@ -94,6 +94,31 @@ type secureBootCertificateImporterProvider struct {
...
+// this call independently. A platform whose equivalent setting is coupled to
+// key-store initialization cannot satisfy this contract and MUST NOT
+// implement this interface.
+//
+// Implementations MUST be idempotent.
+//
+// rebootRequired reports that the change is staged and will not be in effect
+// until the host has been power cycled.
+type SecureBootKeyManagementSetter interface {
+	SetSecureBootKeyManagement(ctx context.Context, enable bool) (rebootRequired bool, err error)
```
joelrebel: The current method name makes it a bit unclear what its purpose is, what do you think of naming this method `AllowCustomSecureBootKeys(ctx, bool)` instead or something similar that clearly indicates its function?

mcanevet: Good call — done in c85a686: `SecureBootKeyManagementSetter`/`SetSecureBootKeyManagement` → `CustomSecureBootKeysAllower`/`AllowCustomSecureBootKeys` (interface, provider implementation, `Client` wrapper, and feature flag all updated together).

joelrebel: thanks

PR #467, opened 2026-09-10
fix(dell): recover instead of failing when a BIOS settings PATCH conflicts

mcanevet:
**Depends on #468** (rebased on top of it) - needs the gofish bump #468 brings in.

## Summary

Dell iDRAC allows only one uncommitted BIOS config job at a time: PATCHing `Bios/Settings` seals it into a job immediately, and any other PATCH to the same settings resource is rejected with `IDRAC.2.14.SYS011` ("Pending configuration values are already committed") until that job is applied at the next reset or cancelled. It's the same job type regardless of which Redfish resource issued the earlier PATCH - confirmed live, a direct PATCH to `ComputerSystem`'s `SecureBoot` resource seals it too (PowerEdge R6715, iDRAC firmware 1.5.3). So two independent BIOS-affecting calls in one boot cycle fail on the second, even though both would apply together at the next reset if submitted as one PATCH.

On that conflict, the recovery reads the pending settings resource, merges the newly requested attributes into whatever is already staged there, cancels the blocking job with a standard Redfish `DELETE`, and resubmits the merged superset - collapsing what would otherwise need two reboots into one. Bounded to a single retry: if the resubmission also conflicts, that's a genuine race to surface, not something to retry blindly. Cancellation refuses to proceed once the job is mid-apply (`Oem.Dell.ActualRunningStartTime` set). Dell's own `DellManager.ClearPending` OEM action is deliberately not used: confirmed live, POSTing to it is itself rejected with the same conflict it would resolve.

Two commits, keeping the provider change separate from the `internal/redfishwrapper` one per the guidance on #460:

1. **`redfishwrapper: add ApplyBiosAttributes and Jobs primitives`** - generic, vendor-blind. `ApplyBiosAttributes` is `SetBiosConfiguration` for already-typed attributes (`schemas.SettingsAttributes`), needed because resubmitting a previously read pending set has to preserve each value's original JSON type - a BIOS attribute registry is typically strict, and resubmitting `"42"` for an integer attribute risks the BMC rejecting the retry. `Jobs` is a vendor-blind `ServiceRoot -> JobService -> Jobs` walk.
2. **`fix(dell): recover instead of failing when a BIOS settings PATCH conflicts`** - the Dell policy. Both the conflict signature and the job-type/cancellation details are Dell OEM specifics, so this lives in `providers/dell` rather than the shared `redfishwrapper`.

## One design choice worth a second opinion

Rather than a separate entry point each BIOS-backed feature has to remember to call, the recovery lives on a `recoveringRedfishClient` that embeds `redfishwrapper.Client` and **shadows** its `SetBiosConfiguration`; `Conn` holds that wrapper.

Upside: every call site keeps reading `c.redfishwrapper.SetBiosConfiguration(...)` and gets recovery without opting in, so this PR doesn't touch each setter, and a *new* BIOS-backed setter can't reach for an unrecovered same-named alternative - there isn't one. The wrapper adds only the recovery; the write itself defers to the embedded client.

The tradeoff, stated plainly: `c.redfishwrapper.SetBiosConfiguration(...)` no longer means quite what a reader assumes; you have to check the field's type to see the override. Go also has no virtual dispatch through embedding, so this covers calls through the wrapper only - naming the embedded field explicitly (`c.redfishwrapper.Client.SetBiosConfiguration`) still bypasses it, and `redfishwrapper`'s own internals would too if they ever called it (they don't today - checked). Both caveats are documented on the type. Happy to switch to an explicit non-shadowing entry point if you'd rather the call sites be self-evident at the cost of each setter opting in.

## Test plan
- `TestSetBiosConfiguration_PendingSettingsConflictRecovery` - happy-path recovery.
- `TestSetBiosConfiguration_PendingSettingsConflictRecoveryIncludesUnchangedAttributes` - guards that `recoverFromPendingSettingsConflict`'s merge still carries forward an already-staged attribute whose value equals the applied one, even though gofish (stmcginnis/gofish#571) may then legitimately omit it from the actual retry PATCH since it's already correctly staged there.
- `TestSetBiosConfiguration_PendingSettingsConflictNoPendingJobFound`, `TestSetBiosConfiguration_PendingSettingsConflictJobAlreadyRunning` - error paths.
- `TestFeatureSetterInheritsConflictRecovery` - guards the shadowing property: a feature setter with no recovery-specific code (`SetHTTPBootURI`) recovers anyway. Verified it fails with the raw conflict if `Conn.redfishwrapper` is retyped back to the bare client.
- Validated live end-to-end on a PowerEdge R6715 (iDRAC firmware 1.5.3): a multi-attribute conflict + recovery cycle through an actual Tinkerbell workflow, confirmed by Redfish query that the resubmitted job held all correct merged values and they applied correctly on reboot.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 new issues), `go test ./...` all pass.


mcanevet:
Closing this without merging. While validating #471, I live-tested the actual `IDRAC.2.14.SYS011` conflict this PR was meant to recover from, across three PowerEdge R6715s (iDRAC 1.20.80.5x), covering:

- every pairwise ordering of the `Bios/Settings` attributes this project's workflows touch (`SecureBoot`, `SecureBootPolicy`, `HttpDev1EnDis`, `HttpDev1Uri`)
- the `BootDeviceSet` boot-override write, a separate staging resource, against pending BIOS attributes, both orders
- same-attribute re-patch (the retry case), including a mind-changing 3-in-a-row chain

The only conflict I could reproduce, in every trial, is a `/SecureBoot` *resource* PATCH issued before a `Bios/Settings` PATCH in the same pending job — never two `Bios/Settings` writes against each other, in any order or combination. #471 removes that one trigger by routing `SetSecureBoot` through the `SecureBoot` BIOS attribute instead of the resource, so nothing in the Dell provider touches the poisoning endpoint anymore. With that in place, I found no remaining scenario where this recovery layer has something to catch.

Happy to reopen if a conflict shows up that doesn't fit this picture — the recovery logic here is sound, I just couldn't find a live case it's still needed for.

PR #472, opened 2026-09-11
fix(dell): don't skip the SecureBootPolicy write when applied state matches

mcanevet:
Stacked on #466 and #468. This PR's diff currently bundles their commits in; it'll shrink to just the last commit once they merge.

`AllowCustomSecureBootKeys` read the attribute's currently applied value and returned early when it already matched what was requested. Confirmed live, that's unsafe: currently-applied state can match while a different value is genuinely pending from an earlier call in the same boot cycle (e.g. a stale pending PATCH left staged by an interrupted, unrelated caller). Skipping the write left that stale pending value in place, silently reverting an explicit request on the next reboot — observed as a certificate import rejected because `SecureBootPolicy` committed as `Standard` despite requesting `Custom` moments earlier.

This is the same class of bug `stmcginnis/gofish#571` (merged, picked up by #468) fixes one layer down, in `SetBiosConfiguration`'s own diff baseline — but #468 alone doesn't cover this case, since the early return here skipped the call to `SetBiosConfiguration` entirely, before #468's fix ever gets a chance to run.

The attribute is still read first, but only to reject platforms that don't expose it at all. The write itself is now unconditional.

Depends on #466 and #468.


mcanevet:
Folded into #466 - this fix is for AllowCustomSecureBootKeys, which #466 itself introduces, so it belongs in that PR rather than as a separate follow-up.

